PR #5, opened 2023-09-27
Add Cargo.lock to version control

Detegr:
It is [recommended](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html) for binary projects to check in the Cargo.lock file to the version control.


kaikalii:
This is true, but the main `uiua` crate is intended to be usable as both a library and a binary.
You should be able to do both
```
cargo install uiua
```
and also
```
cargo add uiua
```
Do you know if there is guidance about the Cargo.lock for crates like this?

Detegr:
According to the Cargo [FAQ](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) Cargo.lock is simply ignored in library crates:

> For libraries the situation is somewhat different. A library is not only used by the library developers, but also any downstream consumers of the library. Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists). This is precisely because a library should not be deterministically recompiled for all users of the library.

PR #17, opened 2023-09-28
Change readme.md to suggest install from GitHub link

nskobelevs:
`cargo` allows installing binaries directly from a git repository.
This provides a better user experience as it allows them to run one command to install the binary and not having to worry about deleting the repo afterwards.


kaikalii:
Oh, I did not know you could do that.
Thanks!

PR #20, opened 2023-09-29
rtl page proofreading

wackbyte:


kaikalii:
You know, I swear I double and triple check these, but there is something is my brain the just ignores the obvious mistakes.
Thanks a lot @wackbyte .

wackbyte:
I'm glad I can help :)

PR #24, opened 2023-09-29
Reduce length of toUpper and toLower

iFreilicht:
Size is reduced by one byte without hurting readability.

Maybe this isn't wanted, though, as the current example shows off that you can clone a value multiple times?


kaikalii:
Hm, I hadn't recognized the equivalence between `f:g..` and `f,g.`.

PR #36, opened 2023-09-30
Update viuer to a version that supports kitty 0.26.0+

bhansconnect:
Currently the latest release of viuer will end up not deleting temp files with kitty. After all temp files are generated viuer will hang. This leads to uiua hanging as well.

Luckily, the fix is simple. Kitty will delete temp files if their name contains `tty-graphics-protocol`. More details in https://github.com/atanunq/viuer/pull/51 (this is a 1 line change to a temp file name)

Since the viuer maintainer has been completely inactive on git since November last year, this PR updates uiua to depend on pull 51. This enables uiua to work on kitty without requiring the viuer maintainer to accept the PR and cut a new release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3345,8 +3345,7 @@ checksum = "49874b5167b65d7193b8aba1567f5c7d93d001cafc34600cee003eda787e483f"
 [[package]]
 name = "viuer"
 version = "0.6.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


kaikalii:
I think depending on a git repo rather than crates.io prevents publishing on crates.io, which means I can't merge this as is.
If viuer is unmaintained, it may be worth it for me to just write image-to-terminal functionality myself.

bhansconnect:
That makes sense. Too bad.

PR #46, opened 2023-10-01
add `n×n identity matrix` uiuism

qexat:
Hi!

I'm an absolute beginner in array languages, but I wanted to give a try and add an uiuism for the identity matrix.

It works, but I'm not sure if it's the most idiomatic code so I'm very open to suggestions!

Thank you in advance for taking your time.


qexat:
I just noticed that other uiuisms work as is, so I'm going to add a test value

kaikalii:
I'm sorry, but there but there is already an identity matrix uiuism :(

qexat:
Oh, I completely missed it, sorry

PR #50, opened 2023-10-01
remove roll and unroll from donut example

bhansconnect:
Just updating to dip. Took a minute to figure it out, but I think this version with dip is easier to follow. Nice to temporarily stash one value while calculating the next. Also rechecked all the math and fixed a bug in how the luminance was calculated.


kaikalii:
Great, thank you!

PR #52, opened 2023-10-01
Added dot product and matrix multiplication

OlegAlexander:
Thank you for creating Uiua! I'd like to add the dot product and matrix multiplication examples.


kaikalii:
Very good!

PR #54, opened 2023-10-01
Update documentation for `under`

shnarazk:
Right?


kaikalii:
No. That's wrong, but my explanation isn't very clear. There's probably a better way to word it.
I'm basically trying to explain why you need to flip the arguments if t on of the is inside g.

shnarazk:
Hmmm, I have no idea. So I'll close this...

PR #59, opened 2023-10-02
Enable passing CLI args to run, eval and watch

iFreilicht:
Right now, passing arguments to the CLI is an error:

```
uiua run test.ua my-arg
error: unexpected argument 'my-arg' found

Usage: uiua run [OPTIONS] [PATH]

For more information, try '--help'.
```

With this PR, these commands all work:

```
$ uiua run test.ua my-arg
[□"test.ua" □"my-arg"]
$ uiua watch my-arg 
Watching for changes... (end with ctrl+C, use `uiua help` to see options)
―――――――――――――――――――――――――――――――――――――――
[□"./test.ua" □"my-arg"]
watching for changes...
$ uiua eval '&args' my-arg
["" □"my-arg"]
```

I decided to use an empty string in the case of "eval" to ensure one could always use `drop 1` to get the actual args, no matter of context.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -79,6 +79,7 @@ fn run() -> UiuaResult {
                     mode,
                     #[cfg(feature = "audio")]
                     audio_options,
+                    args,
                 } => {
                     if let Some(path) = path.or_else(working_file_path) {
                         if !no_format {
@@ -89,6 +90,8 @@ fn run() -> UiuaResult {
                         setup_audio(audio_options);
                         let mut rt = Uiua::with_native_sys()
                             .with_mode(mode)
+                            .with_file_path(&path)
+                            .with_args(args)
                             .print_diagnostics(true);
                         rt.load_file(path)?;
                         for value in rt.take_stack() {
@@ -102,11 +105,13 @@ fn run() -> UiuaResult {
                     code,
                     #[cfg(feature = "audio")]
                     audio_options,
+                    args,
                 } => {
                     #[cfg(feature = "audio")]
                     setup_audio(audio_options);
                     let mut rt = Uiua::with_native_sys()
                         .with_mode(RunMode::Normal)
+                        .with_args(args)
                         .print_diagnostics(true);
                     rt.load_str(&code)?;
                     for value in rt.take_stack() {
@@ -126,8 +131,8 @@ fn run() -> UiuaResult {
                         return Ok(());
                     }
                 }
-                App::Watch { no_format } => {
-                    if let Err(e) = watch(working_file_path().as_deref(), !no_format) {
+                App::Watch { no_format, args } => {
+                    if let Err(e) = watch(working_file_path().as_deref(), !no_format, args) {
                         eprintln!("Error watching file: {e}");
                     }
                 }
@@ -136,7 +141,7 @@ fn run() -> UiuaResult {
             }
         }
         Err(e) if e.kind() == ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand => {
-            if let Err(e) = watch(working_file_path().as_deref(), true) {
+            if let Err(e) = watch(working_file_path().as_deref(), true, Vec::new()) {
                 eprintln!("Error watching file: {e}");
             }
         }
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -89,6 +90,8 @@ fn run() -> UiuaResult {
                         setup_audio(audio_options);
                         let mut rt = Uiua::with_native_sys()
                             .with_mode(mode)
+                            .with_file_path(path.to_str().unwrap().to_owned())
```
kaikalii, line 93: Please use `.to_string_lossy()`. We don't want to panic.

src/main.rs
```diff
@@ -236,8 +241,8 @@ fn watch(initial_path: Option<&Path>, format: bool) -> io::Result<()> {
                                 #[cfg(feature = "audio")]
                                 "--audio-port",
                                 #[cfg(feature = "audio")]
-                                &audio_port,
-                            ])
+                                &audio_port
+                            ]).args(args.clone())
```
kaikalii, line 245: Should be `&[args]`, no need to clone.

src/run.rs
```diff
@@ -206,6 +212,24 @@ impl Uiua {
     pub fn mode(&self) -> RunMode {
         self.mode
     }
+    /// Set the command line arguments
+    pub fn with_args(mut self, args: Vec<String>) -> Self {
+        self.cli_arguments = args;
+        self
+    }
+    /// Get the command line arguments
+    pub fn args(&self) -> Vec<String> {
```
kaikalii, line 221: Should return `&[String]`. No need to clone if they just need to be read.

src/run.rs
```diff
@@ -55,6 +55,10 @@ pub struct Uiua {
     diagnostics: BTreeSet<Diagnostic>,
     /// Print diagnostics as they are encountered
     print_diagnostics: bool,
+    /// Arguments passed from the command line
+    cli_arguments: Vec<String>,
+    /// File that was passed to the interpreter for execution
+    cli_file_path: String,
```
kaikalii, line 61: I think this should be a `PathBuf`

src/run.rs
```diff
@@ -206,6 +212,24 @@ impl Uiua {
...
+    pub fn with_args(mut self, args: Vec<String>) -> Self {
+        self.cli_arguments = args;
+        self
+    }
+    /// Get the command line arguments
+    pub fn args(&self) -> Vec<String> {
+        self.cli_arguments.clone()
+    }
+    /// Set the path of the file that is being executed
+    pub fn with_file_path(mut self, file_path: String) -> Self {
```
kaikalii, line 225: Because `file_path` should be a `PathBuf`, this should take `impl Into<PathBuf>`.

src/run.rs
```diff
@@ -206,6 +212,24 @@ impl Uiua {
...
+        self.cli_file_path = file_path;
+        self
+    }
+    /// Get the path of the file that is being executed
+    pub fn file_path(&self) -> String {
```
kaikalii, line 230: ...and this should return `&Path`

src/sys.rs
```diff
@@ -926,7 +922,8 @@ impl SysOp {
                 env.push(vec![height as f64, width as f64])
             }
             SysOp::Args => {
-                let args = env.backend.args();
+                let mut args = env.args();
+                args.extend(env.args());
```
kaikalii, line 926: Why are the args put in the list twice?

iFreilicht: Whoops, that should've been the file path, of course.

src/run.rs
iFreilicht: Done!

src/run.rs
iFreilicht: Ah nice, that's much more flexible.

src/run.rs
iFreilicht: Makes sense, done!

src/main.rs
iFreilicht: You mean `&args`? But yeah, that makes more sense.

kaikalii: Yep

kaikalii (approved):
Looks good!

PR #65, opened 2023-10-02
Add "--clear" flag to uiua watch

expitau:
Add a flag to clear the terminal when a file is changed, while running "uiua watch". This allows for an (optional) better developer experience

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -333,9 +333,9 @@ fn watch(
             if last_time.elapsed() > Duration::from_millis(100) {
                 if clear {
                     if cfg!(target_os = "windows") {
-                        Command::new("cmd").args(&["/C", "cls"]).status().unwrap();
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -320,6 +329,13 @@ fn watch(initial_path: Option<&Path>, format: bool, args: Vec<String>) -> io::Re
             .last()
         {
             if last_time.elapsed() > Duration::from_millis(100) {
+                if clear {
+                    if cfg!(target_os = "windows") {
+                        Command::new("cmd").args(&["/C", "cls"]).status().unwrap();
```
kaikalii, line 334: `_ = ` to discard the error is probably better here.
You never know what might be available or not

expitau: Changed

kaikalii:
Thanks for your contribution!

PR #66, opened 2023-10-03
`&httpsw`

aspiringLich:
Alright! After some work it's finished.

Some implementation notes:
 - I ended up naming it `&httpsget`.
 - Parameters (?) Arguments (?) Inputs / Outputs (?) (i have no idea what they're called):
   - It takes in an string as an HTTP request, verifies it with the crate `httparse` (to give slightly more helpful error messages than just timing out)
   - Also takes in a socket so that it can time out, and just seemed cleaner that way. Makes `&httpsget` like a more specialized version of `&w`
   - Outputs a string representing an HTTP response.
 - Uses `rustls` to make the HTTPS requests
 - It needs the `Host` header because I'm pretty sure you can't get the DNS / hostname from a TCP socket (which I need for TLS) and this is probably better than taking another parameter.

I couldn't get the website to run so couldn't verify that the documentation isn't misformatted or something (Though the example in the documentation, even if formatted correctly, *will* just throw an error).

<details>
<summary>Terminal log</summary>

```bash
# wont work (ver sad)
[I] ~/p/c/u/site $ cd ..
[I] ~/p/c/uiua $ uiua run examples/http_server.ua
Server started
Error: Cannot join number array and character array
    ╭─[examples/http_server.ua:33:3]
 33 │   ⊂ bytes $ HTTP/1.1 _
    │   ─  
    │       
────╯
  in ⊂                     at examples/http_server.ua:33:3
  in `response`            at examples/http_server.ua:42:15
  in `servererror`         at examples/http_server.ua:45:3
  in !                     at examples/http_server.ua:45:3
  in `handlepageloaderror` at examples/http_server.ua:47:8
  in ⍣                     at examples/http_server.ua:47:8
  in `page`                at examples/http_server.ua:51:8
```

</details>

Also - this command should be possible to run on the browser. You could allow the creation of a dummy socket that you can't write to normally, but can make https requests through. I'll stop here though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -911,7 +911,7 @@ version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "55ac459de2512911e4b674ce33cf20befaba382d05b62b008afc1c8b57cbf181"
 dependencies = [
- "spin",
+ "spin 0.9.8",
 ]
 
 [[package]]
@@ -2385,6 +2385,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "ring"
+version = "0.16.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3053cf52e236a3ed746dfc745aa9cacf1b791d846bdaf412f60a8d7d6e17c8fc"
+dependencies = [
+ "cc",
+ "libc",
+ "once_cell",
+ "spin 0.5.2",
+ "untrusted",
+ "web-sys",
+ "winapi 0.3.9",
+]
+
+[[package]]
 name = "rstml"
 version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


aspiringLich:
I don't know how to mention the issue in the title like the other guy did lol

kaikalii:
This is very cool.
However, why is the request string + tcpsocket handle better than just timeout + url?

aspiringLich:
Ok here’s my reasoning:

1. Consistency - you can make a basic HTTP request with `&w` the same way you make a fancy HTTPS request with `&httpsget`. Might be a moot point for you but I think its neat.
2. For my personal use-case I need custom headers, and, I dunno, someone in the future might want to use it to make api calls or something. Timeout + method and URL + headers + body just feels like way too much, especially if you just want to make a basic https request it would feel very unwieldy imo. Its not *that* bad because you could hide it behind a function but still, 4-5 arguments is too many.
3. At the end of the day it is going through a TCP socket, so I think it makes sense for the function to take one

Concessions (that I will then immediately explain why theyre actually not that bad):

1. It does need people to know HTTP, but honestly if you’re willing to learn a language like Uiua you can learn the HTTP spec too
2. Needing the host name twice is a little redundant, but this could be worked around by storing the hostname with the TCP port and then just dropping in the header (If you really want to)
3. having the big string seems a little against the spirit of this array based language but also idk how to alternative is any more elegant so

I need to sleep soon so I may not see your reply but ultimately, it’s your language so whatever you feel like is best.

I do like my idea better though.

I will implement your idea though if you still think its better.

kaikalii:
I understand your reasoning. Let's keep the symmetry with `&w`.
I want it to be as easy to use as possible though, so I'd like to store the requested hostname with the socket.
Then, I'd like to parse the request string before sending and insert the Host header if it is not present.
I'd like it to insert `HTTP/1.0`, `\r\n`, etc if necessary.

I know that's a bunch, but basically, I want the user to be able to do
```
&httpsget "GET /" &tcpc "example.com:443"
```
but you'll still be able to do whatever more complex requests you'd like.

Also, if you set the request type manually, isn't it more of a `&httpsw` than a `&httpsget`?

Edit: Now that I think of it, all that stuff could be added later without breaking the API, so if you want to merge without it, that's fine with me.

aspiringLich:
Ooh I like that idea. Okay, I’ll implement it later today

aspiringLich:
Actually I kind of like that you have to specify a method. Dont eant to do too much at risk of being unintuitive

aspiringLich:
Alright, that should do it. There might be some small bugs with it still but it seems to work.

kaikalii:
BTW all the checks do is run `cargo test --all`.
If it fails again you should be able to test locally.

aspiringLich:
Yeah, I just forgot to change the outdated examples in the documentation. The latest commit succeeds on `cargo test --all`

kaikalii:
Great work!

PR #68, opened 2023-10-03
Fix breaking behavior in reduce, fold and scan

iFreilicht:
Fold and reduce will not try to append leftovers to their accumulator anymore, while scan will.

I tested the examples and they worked as I expected.


kaikalii:
Excellent!

PR #70, opened 2023-10-04
Implement formatter configuration

Zollerboy1:
This implements configuration of the formatter.

By default, Uiua now searches for a `.fmt.ua` file in the directory of the file that is run, tested, formatted, or watched, and if it isn't found there recursively up the directory tree. If no such file is found in any parent directory, it falls back to the default configuration.

Here is how the `.fmt.ua` file can look like (this is the default configuration):

```
TrailingNewline ← 1             # boolean
CommentSpaceAfterHash ← 1       # boolean
MultilineIndent ← 2             # natural number
CompactMultilineMode ← "auto"   # one of "always", "never", and "auto"
MultilineCompactThreshold ← 10  # natural number
```

The file can contain arbitrary Uiua code (with the exception that syscalls are not allowed at the moment), only the relevant bindings must be constant and of the right type, but there is no requirement to have all or even any of the bindings.

Of course the behavior of the CLI can be changed using the `--format-config` flag, which can be set to either `search-file`, `default`, or a custom path to a configuration file.

The LSP doesn't support formatter configuration yet, but it should be relatively trivial to add.


kaikalii:
This is awesome. Thank you so much.

No I'll have to work on documenting it...

Zollerboy1:
Maybe you can reuse some of the text I wrote above at least. Also, as `FormatConfig` is now defined as a macro similar to all the primitives in `src/primitive/defs.rs`, I hope it should be relatively straightforward to hook the documentation up with that.

On another note: Could you maybe take a look at `formatting` in `src/lsp.rs`? I wasn't able to get the formatter configuration working there, because when I uncomment the stuff there, it seems that VSCode doesn't get a response from the LSP for a formatting request. I tried to debug it but didn't get very far.

kaikalii:
Sure thing

PR #72, opened 2023-10-04
Explain that both can be chained

iFreilicht:
This is what it looks like rendered:

Screenshot 2023-10-04 at 09 58 43

Not sure if using constant is too complicated here, but not putting the result in an array caused the output to be four lines long, which was annoying. I added two spaces to separate the groups of two elements each better.


kaikalii:
I think I'd rather mention that Each chained `both` doubles the number of arguments.

iFreilicht:
Good point. Updated it and fixed the conflicts.

PR #76, opened 2023-10-04
Update uiuisms.rs

44mira:
Added Average, Index Mask ('Where' in APL), GCD

### DIFF
--- a/site/src/uiuisms.rs
+++ b/site/src/uiuisms.rs
@@ -117,6 +117,10 @@ uiuisms!(
     "/↧ [1 4 5 3 2]",
     /// Sort an array
     "⊏⌂. [1 4 5 3 2]",
+    /// Get the average of an array
+    "÷⇉⧻/+ [1 2 3]",
+    // Get the index mask of a bit array
+    "▽∶⇡⧻. [1 0 0 1]",
     /// Get n numbers between 0 and 1 inclusive
     "÷-1∶⇡. 11",
     /// Get n numbers between 0 and 1 exclusive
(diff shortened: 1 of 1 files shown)


site/src/uiuisms.rs
```diff
@@ -117,6 +117,10 @@ uiuisms!(
     "/↧ [1 4 5 3 2]",
     /// Sort an array
     "⊏⌂. [1 4 5 3 2]",
+    /// Get the average of an array
+    "÷⇉⧻/+ [1 2 3]",
+    // Get the index mask of a bit array
+    "▽∶⇡⧻. [1 0 0 1]",
```
kaikalii, line 123: This is actually present further down as "Find the indices of all 1s".
Maybe the description needs to be reworded to make it easier to search for?

site/src/uiuisms.rs
```diff
@@ -173,6 +177,8 @@ uiuisms!(
     r#"▽×⍜'⊟⇌≡\↥.≠@ . "  abc xyz   ""#,
     /// Upscale a matrix
     "▽↯⧻,/÷△.⍉▽↯⧻,30 [0_1 1_0]",
+    /// Get the gcd of the top 2 values on the stack.
+    ";!(|2 ↬≠0.◿,⇉↥↧) 100 10",
```
kaikalii, line 181: I'm hesitant to put a recursive function in here, as I don't think recursion is very "uiuic". The `recur` function is mostly there for completeness.
Is there a more array-y solution?

site/src/uiuisms.rs
44mira: Ah you're correct! My bad.

site/src/uiuisms.rs
44mira: I will look into it but as of the moment, Euclid's method is the only way I know how to get a GCD algorithmically.

PR #80, opened 2023-10-04
Allow internet access during nix build for tests

### LINKED ISSUE #78: Nix build broken by tests which access the internet
```
uiua> &httpsw example:
uiua> &httpsw "GET / " &tcpc "example.com:443"
uiua> thread 'primitive::tests::prim_docs' panicked at '
uiua> Example failed:
uiua> &httpsw "GET / " &tcpc "example.com:443"
uiua> Error: failed to lookup address information: Temporary failure in name resolution
uiua>    ╭─[:1:18]
uiua>  1 │ &httpsw "GET / " &tcpc "example.com:443"
uiua>    │                  ─────
uiua> ───╯
uiua>   in &tcpc at 1:18', src/primitive/mod.rs:897:33
uiua> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
uiua>
uiua>
uiua> failures:
uiua>     primitive::tests::prim_docs
uiua>
uiua> test result: FAILED. 10 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
uiua>
uiua> error: test failed, to rerun pass `--lib`
error: builder for '/nix/store/zxcvmnknvzkdb0n8gwgjn41xdy2kmjh8-uiua-0.0.14.drv' failed with exit code 101
┏━ 1 Errors:
┃ error: builder for '/nix/store/zxcvmnknvzkdb0n8gwgjn41xdy2kmjh8-uiua-0.0.14.drv' failed with exit code 101
┣━ Dependency Graph:
┃ ⚠︎ uiua-0.0.14 failed with exit code 101 after ⏱︎ 5m54s in checkPhase
┣━━━
┗━ ∑︎ ⚠︎ Exited after 1 build failures at 02:39:15 after 5m54s
```
Nix generally requires its builds to be self contained/pure, so it goes to some lengths to prevent builds from depending on external factors. In service of this, by default it disables internet access during the build.

I think there should be some setting to make nix less strict if we want to keep the test. I'll look into it.

sullyj3:
`nix build` was broken by the introduction of tests which access the internet. This gets the build working again. Closes #78

This is a bit of a workaround, ideally I think the build should be hermetic, because I'm a nix ideologue. However, at the same time, I don't want to step on any toes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,9 @@
 {
   description = "A stack-based array programming language";
 
+  # this is to allow tests to access the internet during checkPhase
+  nixConfig.sandbox = "relaxed";
+
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
   inputs.flake-utils.url = "github:numtide/flake-utils";
   inputs.crane = {


kaikalii:
This was addressed in #74.
They said it's probably not a good idea to fetch stuff in tests anyway, and I tend to agree.

sullyj3:
Ah, great, I missed that.

PR #82, opened 2023-10-05
Update URL on run

aprzn123:
Since there have been issues with people accidentally leaving the pad and losing their code, this PR just updates the URL every time the code is run. This matches the behavior of `uiua watch`, as run on save and save on run are effectively equivalent.


iFreilicht:
Just tested the latest version and it works perfectly! Big QoL improvement, thank you!

kaikalii:
Great, thanks!

PR #83, opened 2023-10-05
Added new uiuisms

jtrieb1:
Added matrix power calculation for square matrices, zero matrix construction to match dimensions of a given matrix, and polynomial evaluation at either a scalar or matrix value. The polynomial is represented by an array of coefficients sorted from highest power of "x" to lowest.


kaikalii:
Awesome, thank!

PR #94, opened 2023-10-06
Add support for ouputting fmt result to stdout

Voklen:
Some editors require a formatting command that returns the result to stdout so this PR adds support for that. I made the arguments for it `-O` and `--to-stdout` to match `tar` but we could go for something else.
Sorry if these functions shouldn't be in `main.rs` but I followed where `show_updated_message()` was.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -74,11 +74,9 @@ fn run() -> UiuaResult {
                 )?;
 
                 if let Some(path) = path {
-                    format_file(path, &config)?;
+                    format_single_file(path, &config, formatter_options.stdout)?;
                 } else {
-                    for path in uiua_files() {
-                        format_file(path, &config)?;
-                    }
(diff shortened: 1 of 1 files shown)


gamma-delta:
Helix formatting pipes the file to stdin *and* reads from stdout, so having an ability to read from stdin would be nice here. (I think convention is to use `-` as the filepath?)

Doing `uiua fmt --to-stdout /dev/stdin` works when I run it manually from the command line, but not from Helix for some reason...

This should probably be a new issue. Seems kind of silly to make a PR for a language I've never used though ... (I'm thinking about doing AOC this year in Uiua.)

gamma-delta:
It also looks like even with `--to-stdout` it still mutates the file passed in, which I think is not expected behavior.

Voklen:
@gamma-delta That is indeed unintended, I've just made a pull request to fix that.

PR #95, opened 2023-10-07
binomial expansion (a+b)2

zaxiom13:
(a+b)2 = a2 +2ab +b2 always is true


zaxiom13:
Might it fail occasionally because of floating pointing arithmetic issues? I shouldn't have used random num generator. Going to take this request down.

PR #98, opened 2023-10-07
add a base-10 string parse "uuism"

erplsf:
I think it's a pretty useful one to have in the base "toolbox".

Was not so sure about formatting, so please advise if needed.


kaikalii:
This is a neat implementation, but is it necessary when [parse](https://www.uiua.org/docs/parse) exists?
With some rearranging, it might be a good base `≤10` parser.

erplsf:
Is it new? I spent whole evening looking for it :D

Yes, by changing the 10, any base can be chosen indeed. What rearrangements
you propose?

kaikalii:
All the Uiuisms should be written so that their arguments can be put to the right. So that they can be used like functions.
Your would have to be arranged so that the base and input string were both on the right.
Also, put a space between the arguments and the expression.

erplsf:
Thanks for the instructions - I'll open a new PR with a more idiomatic version.

PR #115, opened 2023-10-10
Fix `@\_`. (was: Implement `@\e` escape sequence)

### LINKED ISSUE #114: Add `\e` escape character
I've been playing around with Uiua for some CLI applications, and I've found that making ANSI escape sequences are a bit hard without having a dedicated `\e` escape sequence.

For example, I have the [following code](https://github.com/dosisod/uordle/blob/b06c8d13066891b3fdca89390081a69621b25cda/uordle.ua#L24-L42):

```
Escape ← + 27 @\0

PrintBanner ← (|0
  ⍥Escape 7
  WordLen
  &p $ uordle: wordle written in uiua.
     $ 
     $ rules:
     $ 
     $ * type words in lower case
     $ * words must be _ characters long
     $ * _[42;30m green  _[0m means correct character is in the right spot
     $ * _[43;30m yellow _[0m means correct character is in the wrong spot
     $ * _[48;5;240m_[30m gray   _[0m means incorrect character
     $ 
     $ type a word to get started
)
```

While this banner is mostly static, I have to inject a bunch of `Escape` values using a format string, which gets hard to manage if I add more escapes later on.

The escape sequence doesn't have to be `\e`, it could be `\x1b` or `\033`, though having a `\xNN` escape sequence would mean you would have 2 ways to make characters (using `+ X @\0` and `"\xNN"`).

What are your thoughts on this? Thanks!

iFreilicht:
~~Fixes `#114`~~

Also fixes a bug that caused the formatter to overwrite `@\_` with `@\u{10ffff}`, which is not valid Uiua.


kaikalii:
Sorry, but this is actually not how I'd like to solve this.
I think it would be better to just let format and multiline strings accept escape character normally.

iFreilicht:
I see, that makes sense. I removed that commit from my branch so you can at least merge the unit tests and fix for `@\_`

iFreilicht:
Oh wait, maybe there's a misunderstanding! Multiline strings already do accept escape characters! The issue is that Uiua doesn't know `\e` specifically:

```
&p $ uordle: wordle written in uiua.
   $ 
   $ rules:
   $ 
   $ * type words in lower case
   $ * words must be _ characters long
   $ * \e[42;30m green  \e[0m means correct character is in the right spot
   $ * \e[43;30m yellow \e[0m means correct character is in the wrong spot
   $ * \e[48;5;240m\e[30m gray \e[0m means incorrect character
   $ 
   $ type a word to get started
```

Would display on the terminal like this:

Screenshot 2023-10-10 at 16 41 32

But right now, Uiua doesn't support that escape sequence, which is why in the example in #114, it has to be inserted via formatting.

Or are you saying you want to implement additional syntax to make writing the sequences above even simpler?

kaikalii:
Shouldn't it be `\0`?

iFreilicht:
No, this is about the Ascii ESC character, a non-prontable control character used to start control sequences, mostly in terminals. See https://en.wikipedia.org/wiki/Escape_character#ASCII_escape_character
and
https://en.wikipedia.org/wiki/ANSI_escape_code

kaikalii:
Ah, I see.
Sounds good then.

iFreilicht:
Haha you were to quick now, I had already removed that commit because I thought you wouldn't merge it :D See #119 now for the actual change.

PR #119, opened 2023-10-10
Add @\e escape sequence for ESC character

### LINKED ISSUE #114: Add `\e` escape character

iFreilicht:
Fixes #114


kaikalii:
I have actually just implemented proper hex character escapes in 065a3bf2a681f3c4cb

iFreilicht:
Ok sure, that works as well :)

PR #120, opened 2023-10-10
add builtin regex function

billy-yoyo:
# Context
This adds the [Rust Regex](https://docs.rs/regex/latest/regex/index.html) package as a dependency and adds a simple `regex` function (without a glyph alias) which executes a pattern and returns a list of boxed strings representing each of the matching groups.

I also added a hashmap cache to the Uiua environment, which we use as a pattern cache.

Let me know if you'd like any changes.


billy-yoyo:
actually just realized this might not handle error situations very well, will add some more tests

kaikalii:
Great work!
I'd like to request one change.
Rather than having the cache as part of the `Uiua` struct, can you make it a thread-local?.
That way calling `regex` won't have to lock anything.

billy-yoyo:
Will do, probs won't have time tonight but maybe tomorrow

billy-yoyo:
@kaikalii changed the cache to use a thread local and also added some recovery code (invalid pattern, 0 matches) so it should be more stable now.

billy-yoyo:
That's slightly annoying, the feature it's erroring for because it's unstable has been all but approved: https://github.com/rust-lang/rust/issues/92122

Not sure if there's a way to disable this error as it's not really important based on that rust Lang pr

kaikalii:
Why not just use `.with(...)`?

billy-yoyo:
@kaikalii changed with_borrow_mut to with followed by a borrow_mut

Issue #60, opened 2023-10-02
a way to convert glyphs back into name


goyalyashpal:
### what
a way to convert glyphs back into name

### use cases
* super helpful in migration from breaking changes in glyph designation like in #58 
* documentation or tutorial material where showing names make more sense 

e.g. for "showing names more sensible":
* recommended font [dejavu] not suitable or not to the liking - say for print media like books
* in a beginner tutorial - where you are introducing this language to newcomers & you _want_ to start with names
* in accessibility critical usage: i.e. for people with eye sight problems
* where you (content creator/book author) can't expect reader to remember specifics of any language as you are covering huge breadth of prog langs/paradigms
* better searchability of keywords

[dejavu]: https://www.uiua.org/docs/install

kaikalii:
If someone would like to implement this, it can be added to the new formatter configuration.
This is a common request, so I'll try to get to it if I can.

kaikalii:
My philosophy on this has changed.
Having unformatted code is not something I want to encourage.
Reasoning [here](https://github.com/uiua-lang/uiua/pull/121#issuecomment-1756502921)

PR #121, opened 2023-10-11
Add formatter configuration option for disabling glyphs

billy-yoyo:
# Context
Implementing this issue: https://github.com/uiua-lang/uiua/issues/60

Adds a `DisableGlyphs` flag to the `.fmt.ua` format configuration. If this is set to true, all glyphs will be converted back to their ascii representations. The formatter will also try to add spacing to make the names clearer when this is enabled.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -19,6 +19,7 @@ use crate::{
     parse::parse,
     value::Value,
     SysBackend, Uiua, UiuaError, UiuaResult,
+    primitive::PrimitiveFormatOptions
 };
 
 // For now disallow any syscalls in the format config file.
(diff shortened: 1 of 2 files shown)


kaikalii:
What is the motivation for this?

billy-yoyo:
The issue I linked - if people find keeping track of what the glyphs mean confusing this allows them to write functions by name and keep track of them better.

kaikalii:
I think writing code this way, and having unformatted code in general, is actually something I don't want to encourage.
Having unformatted names leaves code open to breakage when new built-in functions are added, and not reading the glyphs could prevent you from having to learn them.
I'm sorry I left that issue open.

billy-yoyo:
Fair enough, no problem!

PR #127, opened 2023-10-13
Add arithmetic mean to uiuisms.rs

cabralski:
i hope this implementation is simple and good enough. what an incredible language!


kaikalii:
Yes, this is a great way to do it!

PR #140, opened 2023-10-16
Create dependabot.yml

shnarazk:
Hi,
Is there any reason not to use Dependabot? It automatically creates pull requests to update package dependencies, like this. 

スクリーンショット 2023-10-15 20 47 38

So it would be helpful.

- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates


kaikalii:
Sure, why not

PR #152, opened 2023-10-17
Add web backend for &runc and &runi

expitau:
Adds javascript eval as a command backend for run_command_capture and run_command_inherit.

Example usage
```
&runi {"alert" "'hello, world'"} # 0
⊙;;&runc "3+7" # 10
⊙;;&runc "Date.now()" # "1697515821857"
```

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -34,6 +34,17 @@
       }
     }(window.location))
   </script>
+  <script>
+    // This is for calling javascript with &runi
+    function run_js(f, args) {
+      try {
+        output = eval(f)
+        return [0, '' + output, null]
+      } catch (e) {
+        return [1, null, e.message]
+      }
+    }
+  </script>
   <!-- End Single Page Apps for GitHub Pages -->
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
--- a/site/src/backend.rs
+++ b/site/src/backend.rs
@@ -10,6 +10,12 @@ use std::{
 
 use leptos::*;
 use uiua::{value::Value, DiagnosticKind, Handle, SysBackend, Uiua, UiuaError, UiuaResult};
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern "C" {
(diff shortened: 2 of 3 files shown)


kaikalii:
Can I ask what your motivating case is for this? I'm inclined to reject it, but if your reason is cool enough, maybe it will be okay.
I only fear that people will showcase Uiua code where all the heavy lifting is being done in Javascript, which is undesirable.
*If* this is merged, it should use `js_sys::eval` rather than manually `wasm_bindgen`ing a function.

edhebi:
This also feels like a particularly arbitrary choice when running uiua natively

expitau:
The idea behind this pull request is to provide more options for interacting with a web-based environment, especially before some protocols are properly implemented (ie. fetch). This allows calling any of the countless native [web apis](https://developer.mozilla.org/en-US/docs/Web/API) available to us in Javascript. Some of my ideas include
- Prompting for user input
- Making fetch requests (bad apple in a shareable link?)
- Creating dom element trees
- Interacting with the clipboard
- Device sensors & haptics
- File uploads
- Interacting with webcams and microphones

And probably more I haven't even thought of. I understand your concern regarding people writing more javascript than uiua in a showcase, so I might propose removing Javascript evaluation by default and only allowing for function calls on the window object. (ie only allowing `&runi {"alert", "'hello world'"}` and removing the ability to do `&runi "alert('hello world')"`) to discourage this use. (though note it can still be deliberately bypassed with `&runi {"eval" "1+2"}`)

js_sys::eval does not allow for interacting with native web apis, and runs code in a sandboxed environment, defeating the purpose for this entirely. That is why I chose to implement this with wasm_bindgen. I do admit that the implementation is a little scuffed as I'm trying to fold it into the existing definition for &runc, so status and stderr are effectively unused.

kaikalii:
Thinking on it more, I realize that this would allow the execution of arbitrary Javascript from a link that *looks* like its just Uiua code. Especially in the case of fetch requests, I do not want to allow this. People share links to the pad a lot, and I don't want to open the door to bad actors, so I am closing this.

expitau:
Rip :headstone:

PR #175, opened 2023-10-23
Simple REPL

twolodzko:
I found the uiua's learning curve quite steep. I was missing REPL to effortlessly run, modify code, and see the results. I assumed that if there exists a language server, creating a REPL should not be hard. This proved correct. In this PR I am proposing a very simple REPL that could be useful for others as well.

The read-eval-print loop is initialized with the `uiua repl` command by starting the `uiua::Uiua` server. Then, it loops by following the steps below:
* Prompt the user for the code `»`. It assumes that the code is inputted as a single line. Multiline input is not supported. 
* If using the `-O` flag (as other commands), the code is formatted and printed (prefixed by `↪`).
* Next, the code is evaluated by the server and the output stack is printed. Each element of the stack is printed prefixed by the "therefore" symbol `∴` (which also looks like a stack of three dots). 
* On any error, the messages are printed prefixed by `⚠` and the loop continues.

I didn't test this on any non-Linux machines. I also didn't update any documentation, because I'm not sure if you would like to actively promote such an early-stage feature.


kaikalii:
Great work!

PR #188, opened 2023-10-25
Fix typos and grammar in `other.rs`

citizen428:
While reading https://www.uiua.org/docs/rtl I noticed a grammar mistake. I then noticed some other things in the same file and decided to fix those too.


kaikalii:
My terrible typing strikes again.
Thanks a lot!

PR #200, opened 2023-10-28
Replace TitleCase with CamelCase

julesreid:
Title case uses spaces as separators between words, whereas camel case uses capital letters.  Refer to the Wikipedia article for more information.


kaikalii:
Maybe "PascalCase" would be more appropriate?
Since camel case can also be "camelCase".

julesreid:
Yes, I was looking for a way to specify that.  It helps to fully read the documentation.  I'll resubmit using "PascalCase" but referring to the same Wikipedia article.

PR #208, opened 2023-10-30
Mention cmd as modifier in uiuapad help

memark:
Mention cmd as well as ctrl in uiuapad help, since cmd is the correct modifier on MacOS.

### DIFF
--- a/site/src/editor/mod.rs
+++ b/site/src/editor/mod.rs
@@ -1012,12 +1012,12 @@ pub fn Editor<'a>(
                     <button
                         class="info-button"
                         data-title="\
-shift Enter   - Run + Format
- ctrl /       - Toggle line comment
-  alt Up/Down - Swap lines
-shift Delete  - Delete lines
- ctrl Z       - Undo
- ctrl Y       - Redo"
+   shift Enter   - Run + Format
+ctrl/cmd /       - Toggle line comment
+     alt Up/Down - Swap lines
+   shift Delete  - Delete lines
+ctrl/cmd Z       - Undo
+ctrl/cmd Y       - Redo"
                         disabled>
                         "🛈"
                     </button>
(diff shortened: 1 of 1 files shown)


site/src/editor/mod.rs
```diff
@@ -1012,12 +1012,12 @@ pub fn Editor<'a>(
                     <button
                         class="info-button"
                         data-title="\
-shift Enter   - Run + Format
- ctrl /       - Toggle line comment
-  alt Up/Down - Swap lines
-shift Delete  - Delete lines
- ctrl Z       - Undo
- ctrl Y       - Redo"
+   shift Enter   - Run + Format
```
kaikalii, line 1015: This will have to go on the previous line, since `\` eats all subsequent whitespace.

memark: > since \ eats all subsequent whitespace
Does it do that only on the first line? Since other lines already start with whitespace (like "  alt").

site/src/editor/mod.rs
```diff
@@ -1026,7 +1026,7 @@ shift Delete  - Delete lines
                     <div id="code-area">
                         <div id={glyph_doc_id} class="glyph-doc" style="display: none">
                             { move || glyph_doc.get() }
-                            <div class="glyph-doc-ctrl-click">"Shift+click for more info (Ctrl+click for new tab)"</div>
+                            <div class="glyph-doc-ctrl-click">"Shift+click for more info (Ctrl/Cmd+click for new tab)"</div>
```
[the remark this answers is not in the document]
memark: That's a good idea. (Actually it's what the Apple guidelines say as well.)

site/src/editor/mod.rs
kaikalii: Yes, only the first

site/src/editor/mod.rs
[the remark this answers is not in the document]
memark: Done

site/src/editor/mod.rs
memark: Done

site/src/editor/mod.rs
```diff
@@ -1011,13 +1011,12 @@ pub fn Editor<'a>(
...
                     <button
                         class="info-button"
-                        data-title="\
-shift Enter   - Run + Format
- ctrl /       - Toggle line comment
-  alt Up/Down - Swap lines
-shift Delete  - Delete lines
- ctrl Z       - Undo
- ctrl Y       - Redo"
+                        data-title="\ shift Enter   - Run + Format
```
kaikalii, line 1014: The `\` is no longer needed here. You would be trying to escape the space.

memark: LOL. Gotcha.

PR #215, opened 2023-10-31
Using rustyline in REPL for better user experience

twolodzko:
Thanks to [rustyline](https://github.com/kkawakam/rustyline) REPL can remember the history of inputs and will recognize standard keyboard shortcuts like Ctrl+A or Ctrl+K. It should be platform-independent as assured by the creators of the package.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -436,6 +436,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "702fc72eb24e5a1e48ce58027a675bc24edd52096d5397d4aea7c6dd9eca0bd1"
 
 [[package]]
+name = "clipboard-win"
+version = "4.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7191c27c2357d9b7ef96baac1773290d4ca63b24205b82a3fd8a0637afcf0362"
+dependencies = [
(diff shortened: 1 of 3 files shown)


twolodzko:
@iFreilicht this is the PR with REPL improvements.

memark:
This is something I would really like. I've been pressing arrow up too many times to remember already.

kaikalii:
I'm having a problem in Powershell where it insert extra space after the `»` which disappears after a backspace, but the cursor stays shifted to the right.

twolodzko:
@kaikalii I'm afraid "it works for me" and I don't have access to a Windows machine, so can't reproduce this.

kaikalii:
Looks like it was a problem with the ANSI formatting.
I fixed it locally, but I can't seem to push to update the PR.

memark:
@twolodzko Powershell is also available for Linux and MacOS if you want to test.
https://github.com/PowerShell/PowerShell

twolodzko:
That's strange @kaikalii because I allowed for edits in GitHub UI. Maybe you could try again or generate a git patch that I could apply?

twolodzko:
Btw, I followed @memark's suggestion and installed PowerShell for Linux and it seems to work correctly. So it seems to be a Windows (?) issue.

PR #218, opened 2023-11-02
Update Arrays documentation

nashamri:
Hello there 👋

Thanks for this amazing project, I've been following and enjoying it's progress recently. I'm new to uiua and to array languages in general, and this is a little contribution that I came across while I was reading the Arrays tutorial. In one of the examples, it shows a 2x3 array and both the `length` and `rank` of that array is 2. I thought this might be a little bit confusing to new comers. So I made the array a bit bigger to make the idea clearer.

Keep up the good work!


kaikalii:
Thanks, good idea!

PR #227, opened 2023-11-05
Update Uiua386 to 0.1.0

gifti258:
Glyph overview:
Screenshot_20231105_170755

* Add lots of glyphs, including those added since the last update
* Redraw letters
* Hinting

There's still a lot I want to improve but I didn't want to make you wait any longer.


kaikalii:
Amazing work!
Thank you so much.
I'll put a not in the install docs about getting the font as well.

PR #237, opened 2023-11-07
use "array" instead of "list" in uiuisms

memark:
"Arrays" in Uiua are called just that (not "lists"). Rename a few uiuisms to match that.


kaikalii:
A list is a name for a rank-1 array.
This is in the docs.

memark:
Fair enough, my bad. After some searching I did find it mentioned in the Language Tour.

PR #277, opened 2023-12-02
Don't mutate file with `--to-stdout`

Voklen:
This stops the Uiua file being mutated if `--to-stdout` is used with `uiua fmt` as mentioned in #94


kaikalii:
Cool, thanks!

PR #278, opened 2023-12-02
Update regex behavior

pythongirl325:
Update `regex` function to be more consistent. It will return a 2d array with one column for each capture group and one row for each match.

### DIFF
--- a/src/primitive/mod.rs
+++ b/src/primitive/mod.rs
@@ -808,23 +808,19 @@ fn regex(env: &mut Uiua) -> UiuaResult {
                 Regex::new(&pattern).map_err(|e| env.error(format!("Invalid pattern: {}", e)))?;
             cache.entry(pattern.clone()).or_insert(regex.clone())
         };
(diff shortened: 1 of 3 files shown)


src/primitive/mod.rs
```diff
@@ -808,22 +808,19 @@ fn regex(env: &mut Uiua) -> UiuaResult {
...
-                        .flatten()
-                        .map(|m| Boxed(Value::from(m.as_str())))
-                        .collect()
-                })
-                .unwrap_or_default()
-        };
+                .collect();
+            matches.append(row.into(), env)?;
+        }
+
```
kaikalii, line 823: I think `matches` could be an `Array<Boxed>` rather than a `Value`.
Then it could be turned into a `Value` once everything is appended.

pythongirl325: The `append` function on `Array` is private. Could it be made public or public inside the crate?

PR #281, opened 2023-12-04
Rename lsp token types for compatability with vim highlighting groups

xynxynxyn:
Vim doesn't support hyphens in highlighting groups. Replacing them with underscores fixes the issue.


kaikalii:
Didn't know that. Thanks!

PR #311, opened 2023-12-13
Fix running the test files

jonathanperret:
### Observed problem

Since the `bytecode` branch was merged, the interpreter's test suite (`.ua` files in `tests/`) is no longer run by `cargo test`.

This is easy to see by changing any of the assertions in a `.ua` file and running `cargo test`: the tests are still reported as passing.

### Proposed solution

In the `suite()` test case, use the new `run_file` instead of `load_file` (which used to run the Uiua code but no longer does since compilation and execution were separated).

That function also scanned the diagnostics from the Uiua compiler using `take_diagnostics()` but this function is documented to return an empty list if `print_diagnostics(true)` has been called, which `suite()` used to do.

I changed it to call `print_diagnostics(false)`, and report a failure if any diagnostic more severe than `Advice` is encountered in the test files. I would have stopped on any diagnostic but there are six instances of style advices triggered by the current test files — these could be changed instead.

Finally, to get the tests to pass I had to add a missing expected value in a recently added assertion in `loops.ua`.


kaikalii:
Thanks. Big oversight on my part

kaikalii:
Fortunately I ran them manually often enough that they seem to still pass 😅

PR #321, opened 2023-12-18
Got it to compile on Windows 11 ARM64

axbannaz:
running `cargo update` and changing the `term_size` crate to v0.3.2 (this crate is not maintained any more)
combination of the two fixes compile under Windows 11 on Arm64. `cargo test` pass as well.


axbannaz:
closing in favor of #323

PR #323, opened 2023-12-19
Replace term_size with terminal_size, works on Windows 11 Arm64

axbannaz:
Replacing unmaintained term_size crate with terminal_size.
provides the added benefit of working on Windows 11 ARM64

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -703,7 +703,7 @@ dependencies = [
  "parking_lot",
  "signal-hook",
  "signal-hook-mio",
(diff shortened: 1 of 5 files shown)


kaikalii:
Sorry, I have to revert this, as it breaks the build on wasm.
I'll change the actions so they catch that.

axbannaz:
how about a less evasive change.. 
https://github.com/uiua-lang/uiua/compare/main...axbannaz:uiua:update_term_size
downgrading term_size to latest non-beta.

PR #332, opened 2023-12-26
Add workflow for GitHub release builds

Vap0r1ze:
Merry Christmas 🎄❄

This PR adds a workflow that does the following each time a versioned tag is created:
- Ensures the tag matches the uiua crate version
- Creates a release and pulls the respective `changelog.md` section as the description
- Builds and uploads binaries for multiple targets[^1] on the release

If you also wanted this to work for `cargo publish`, let me know the details of what goes into manually publishing a version.

[^1]: This PR adds builds for the following targets: `x86_64-apple-darwin`, `aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, `x86_64-unknown-linux-musl`, `x86_64-pc-windows-msvc`, `aarch64-pc-windows-msvc`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,15 +49,15 @@ jobs:
 
       # Install dependencies
       - if: matrix.target == 'armv7-unknown-linux-gnueabihf'
-        run: apt install -y gcc-arm-linux-gnueabihf
+        run: sudo apt-get install -y gcc-arm-linux-gnueabihf
       - if: matrix.target == 'aarch64-unknown-linux-gnu'
-        run: apt install -y gcc-aarch64-linux-gnu
+        run: sudo apt-get install -y gcc-aarch64-linux-gnu
       - if: matrix.target == 'x86_64-unknown-linux-musl'
-        run: apt install -y musl-tools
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      # Install dependencies
+      - if: matrix.target == 'x86_64-unknown-linux-musl'
+        run: sudo apt-get install -y musl-tools
+
+      # Build
+      - name: Install target
+        run: rustup target add ${{ matrix.target }}
+      - name: Build
+        run: cargo build --bin uiua --release --target ${{ matrix.target }} --verbose
```
kaikalii, line 60: I'd like release builds to have the `audio` feature enabled.
This usually requires some dependencies to be installed on Linux targets.

Vap0r1ze: I had to remove the `x86_64-unknown-linux-musl` target because for some reason [the linker dies](https://github.com/Vap0r1ze/uiua/actions/runs/7332671620/job/19967152399), otherwise added!

kaikalii:
Awesome, thanks!

PR #335, opened 2023-12-28
Allow virtual term input in raw mode

Vap0r1ze:
This PR adds support for virtual terminal input (e.g: `\x1b[B` for DownArrow) when reading from stdin in raw mode.
It does this by moving from crossterm to a [(smaller) crate](https://codeberg.org/twink/rawrrr) specifically for cross-platform tty raw mode. It works almost exactly like crossterm, but also uses the virtual terminal mode flag.

This also fixes a small bug with the uiua watch-runner where raw mode does not get disabled after the program exits, because `disable_raw_mode()` only worked after `enable_raw_mode()` was called at least once, which the main process never does, only the child process (even though they share the same stdio).

Motivation: I wanted to be handle arrow keys so I could make Tetris :^)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2541,6 +2541,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2ff9a1f06a88b01621b7ae906ef0211290d1c8a168a15542486a8f61c0833b9"
 
 [[package]]
+name = "rawrrr"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


Vap0r1ze:
One possibly weird behaviour about the current state of this PR is that raw mode causes Ctrl+C to be passed down to the uiua program rather than being caught by the main process. Let me know if this is desired or if I should change this in some way because it can be annoying

kaikalii:
This is the current behavior, so I think it's fine.

kaikalii:
Thanks!

PR #351, opened 2024-01-12
Support long unicode escape sequences

jonathanperret:
## Problem

Uiua doesn't currently support Unicode escape sequences for codepoints above `U+FFFF`.

## Proposed solution

Use the same syntax as Rust, i.e. `\u{nnnnnn}` with up to 6 digits.


kaikalii:
Very cool, thanks!
I'll have to add a note to the docs about this.

jonathanperret:
Thanks for merging. I’ll try to remember to update the docs as well next time!

PR #378, opened 2024-02-05
add numeric and letter check uiuisms

JohnBakhmat:
Added uiuisms 
 * to check if character is numeric `[0-9]`. [Numberic demo](https://uiua.org/pad?src=0_8_0__SXNJbnQg4oaQIC894oqC4oqDKOKJpUAwKSjiiaRAOSkKCklzSW50ICIwIgpJc0ludCAiOSIKSXNJbnQgIjQiCklzSW50ICJBIgpJc0ludCAiYSIKSXNJbnQgImIiCklzSW50ICIpIgpJc0ludCAiXyIK)
 * to check if character is english letter `[a-zA-Z]`. [Letter check demo](https://uiua.org/pad?src=0_8_0__SXNMZXR0ZXIg4oaQID0zLyviioLiioLiioLiioPiioPiioMo4omlQEEpKOKJpEBaKSjiiaVAYSko4omkQHopCgpJc0xldHRlciAiMCIKSXNMZXR0ZXIgIjkiCklzTGV0dGVyICI0IgpJc0xldHRlciAiQSIKSXNMZXR0ZXIgImEiCklzTGV0dGVyICJiIgpJc0xldHRlciAiKSIKSXNMZXR0ZXIgIl8iCg==)

In numeric solution i used `/=` as a `AND` operator coz if char is number, the function will return [1,1], if its char code is smaller then `@0` it will return [0,1] and if its higher then `@9` it will be `[1,0]` so there's no way to get `[0,0]`

In letter solution we compare to 3 because for example letter `@b` will return [1 0 1 1] and we need to have `(1 and 0) or (1 and 1)`. So basically if we have 3 `1` then at least one condition is met, if we have 2 `1` neither condition is met. 

There is probably better solution without magic

### DIFF
--- a/site/src/uiuisms.rs
+++ b/site/src/uiuisms.rs
@@ -138,7 +138,7 @@ uiuisms!(
     /// Check if character is numberic [0-9]
     "/=⊂⊃(≥@0)(≤@9)",
     /// Check if character is letter [a-zA-Z]
-    "≥3/+⊂⊂⊂⊃⊃⊃(≥@a)(≤@z)(≥@A)(≤@Z)",
+    "=3/+⊂⊂⊂⊃⊃⊃(≥@a)(≤@z)(≥@A)(≤@Z)",
     /// Check if an array is a palindrome
     r#"≍⇌. "racecar""#,
     /// Convert a number to a string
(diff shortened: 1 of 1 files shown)


JohnBakhmat:
:o Oh yeah that makes sense. I think it also needs `⊢` in the end so we go from `[0]` to `0`

JohnBakhmat:
Oh yeah if we call it on string with length > 1 it will return results for each char, that makes sense, removing `first` now

PR #387, opened 2024-02-10
Get rand seed from OS

hatoo:
Hi, using the current time as a random seed is not the best way to initialize random.

This PR makes `uiua` initialize random more sophisticatedly and should work in wasm too.


kaikalii:
Thanks!

PR #388, opened 2024-02-10
Create a repr primitive

teeth-eator:
Example:
```py
{1 2 "asdf" @a [4 5]}
{[[][][]]}
{□□.□.↯3_3 0}
□[[][][]]
[[][]]
□1
□@a
□"asdf"
□[4 4]
ℂ 1 2
[1 2 3]
+0.1 0.2
NaN
[[1 23 4135][4213 1 23]]
[]
[[[1 2][3 4]][[5 6][7 8]]]
"asdf"
@a
["sdf" "qwe" "abc"]
[""]
[{[[1 2][2 3]][3 4]}{[[1 2][2 3]]{[[1 2][2 3]][3 4]}}]
π
↯ 2_2_2 0
dump(&p repr.)
```
produces
```py
{1
 2
 "asdf"
 @a
 [4 5]}
{[[]
  []
  []]}
{□□□[[0 0 0]
     [0 0 0]
     [0 0 0]]
 □[[0 0 0]
   [0 0 0]
   [0 0 0]]
 [[0 0 0]
  [0 0 0]
  [0 0 0]]}
□[[]
  []
  []]
[[]
 []]
□1
□@a
□"asdf"
□[4 4]
ℂ1 2
[1 2 3]
0.30000000000000004
NaN
[[1 23 4135]
 [4213 1 23]]
[]
[[[1 2]
  [3 4]]
 [[5 6]
  [7 8]]]
"asdf"
@a
["sdf"
 "qwe"
 "abc"]
[""]
[{[[1 2]
   [2 3]]
  [3 4]}
 {[[1 2]
   [2 3]]
  {[[1 2]
    [2 3]]
   [3 4]}}]
3.141592653589793
[[[0 0]
  [0 0]]
 [[0 0]
  [0 0]]]
  ```
  closes https://github.com/uiua-lang/uiua/issues/384

### CODE REVIEW

File: src/array.rs [hunk 1]
```diff
@@ -569,6 +569,18 @@ pub trait ArrayValue:
...
+        ("[", "]")
+    }
+    /// String representation that can be read by the interpreter
+    fn debug_string(&self, _depth: usize, _prefix: bool) -> String {
+        self.to_string()
+    }
+    /// Separator for repr
+    fn debug_sep() -> &'static str {
+        Self::format_sep()
+    }
```
kaikalii, line 583: I'd like to separate these functions from the main `ArrayValue` trait.
Can you make a super trait, i.e. `trait DebugArrayValue: ArrayValue` please?
Put those implementaiton at the bottom of the file.

src/array.rs
```diff
@@ -773,3 +809,65 @@ impl<'a> fmt::Display for FormatShape<'a> {
...
+/// Convert value into a string that can be understood by the interpreter
+pub(crate) fn dbg_value (value: &Value, depth: Option<usize>, prefix: bool) -> String {
+    value.generic_ref(
+                |a| dbg_array::<f64>(a, depth, prefix),
+                |a| dbg_array::<u8>(a, depth, prefix),
+                |a| dbg_array::<crate::complex::Complex>(a, depth, prefix),
+                |a| dbg_array::<char>(a, depth, prefix),
+                |a| dbg_array::<Boxed>(a, depth, prefix),
+                )
+}
```
kaikalii, line 822: Please apply `rustfmt` to all files.

kaikalii (commented):
I think you may have said, but can you clarify: What is the motivating case for this function?

src/array.rs
teeth-eator: done

src/array.rs
teeth-eator: done

kaikalii:
I'd like to mark this as experimental for now. You can do this in `Primitive::is_experimental`.
Can you add some tests to `/tests/units.ua`, near the bottom with the other experimental stuff?
Also, can you add a note about `repr` to the changelog?
Thanks!

teeth-eator:
> I think you may have said, but can you clarify: What is the motivating case for this function?

I often find myself copying output back into my code to more easily modify or reproduce it, and the output format being different from the format used by the language makes this annoying. `repr` bridges that gap by providing a more formal and portable way to print a value.

src/primitive/defs.rs
```diff
@@ -2104,6 +2104,27 @@ primitive!(
     /// ex: 1_2_3 4 5_6_7
     ///   : dump⊢
     (0(0)[1], Dump, Stack, "dump"),
+    /// Stringify any value
```
kaikalii, line 2107: Maybe "convert a value to its code representation"?

teeth-eator:
> I'd like to mark this as experimental for now. You can do this in `Primitive::is_experimental`. Can you add some tests to `/tests/units.ua`, near the bottom with the other experimental stuff? Also, can you add a note about `repr` to the changelog? Thanks!

Done

src/primitive/defs.rs
```diff
@@ -2104,6 +2104,27 @@ primitive!(
...
+    ///   : &p repr ↯2_2_2 0
+    /// ex: # Experimental!
+    ///   : &p repr {"Uiua" @A [1 2 3] □4}
+    ///
+    /// Append commas to whitespace for a more traditional look.
+    /// ex: # Experimental!
+    ///   : ↯2_2_3⇡12
+    ///   : &p ⍜⊜□⎏(⊂@,)∊," \n" repr
+    ///
+    /// [map]s need to me unmapped before calling [repr].
```
kaikalii, line 2123: Can you check for the presence of the `map_len` metadata field, then generate a `map` call appropriately?

teeth-eator: Okay. Automatically unmapping is doable, but putting `map` calls into strings gets very weird because the precedence changes based on the context. I'll see what I can do

teeth-eator: > Can you check for the presence of the `map_len` metadata field, then generate a `map` call appropriately?

added code to clear out empty rows for maps, but I'm not sure how to go about printing a `map` call in a way that doesn't depend on line execution order.
here's two ways I could think of:
изображение

teeth-eator: [this](https://www.uiua.org/pad?src=0_8_0__IyBFeHBlcmltZW50YWwhCgojIE5vcm1hbCBleGVjdXRvbiBvcmRlcgp7IkNhcm9sIgogIkFsaWNlIgogIkJvYiJ9CltbNCA1XQogWzMgOF0KIFsxMiAyXV0KbWFwCgojIEFycmF5IGV4ZWN1dGlvbiBvcmRlcgriiqJbCiAgbWFwCiAgW1s0IDVdCiAgIFszIDhdCiAgIFsxMiAyXV0KICB7IkNhcm9sIgogICAiQWxpY2UiCiAgICJCb2IifV0K) is what the problem is, to be specific

kaikalii: > or I could print it as a box array: изображение this doesn't look weird and still works with `insert` that's how it works for now

I think this is fine.

kaikalii:
This looks great. Thank you!

PR #397, opened 2024-02-13
Support LSP completion for Primitives

hatoo:
image

Support LSP completion only for `Primitives` (for now)


kaikalii:
This is awesome, thanks!

PR #399, opened 2024-02-18
Add support for `‼` in custom modifier names

### LINKED ISSUE #398: Suggestion: Format 2+adic custom modifiers using `‼` (double exclamation point)
This was mentioned on the Discord server a while ago and I like the idea. You would be able to type a custom modifier name with normal exclamation points, and the formatter would handle converting `Example!!` -> `Example‼`. Even-adic modifiers would use half as many `‼`s, and odd-adic modifiers would use `‼`s followed by one `!`.

Omnikar:
Closes #398. Custom modifier names are now always parsed and formatted as their canonicalized form with ⌊n/2⌋ `‼` characters followed by one `!` if n is odd.


kaikalii:
Cool!

PR #403, opened 2024-02-19
Revise several uiuisms

Omnikar:
Functionality changes:
- Cross product had incorrect signs before; this is now fixed.
- Pad an array with 0s now works on arrays of any rank.
- Powerset outputs in a different order.

All other changes are style/conciseness only.

### DIFF
--- a/site/src/uiuisms.rs
+++ b/site/src/uiuisms.rs
@@ -148,7 +148,7 @@ uiuisms!(
     /// Check if an array is a palindrome
     r#"≍⇌. "racecar""#,
     /// Convert a number to a string
-    r#"$"_" 17"#,
+    r#"°⋕ 17"#,
     /// Convert a string to a number
     r#"⋕ "42""#,
     /// Convert a string to a list of code points
@@ -194,11 +194,11 @@ uiuisms!(
     /// Filter by a fixed predicate
     "▽ =0◿2 . ⇡10",
     /// Find the most common row in an array
-    r#"⊏⊢⍖°⊚⊛:◴. "Hello World!""#,
+    r#"⊡⊢⍖°⊚⊛:◴. "Hello World!""#,
     /// Convert a string to uppercase
-    r#"-×32×≥@a,≤@z. "These are Words""#,
+    r#"⌵ "These are Words""#,
     /// Convert a string to lowercase
-    r#"+×32×≥@A,≤@Z. "These are Words""#,
+    r#"¯⌵ "These are Words""#,
     /// Check if a string is in a list of strings
     r#"∊□ "uiua" {"apl" "bqn" "uiua"}"#,
     /// Trim leading whitespace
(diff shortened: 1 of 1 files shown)


site/src/uiuisms.rs
```diff
@@ -228,15 +228,15 @@ uiuisms!(
...
+    "1_2_3 4_5_6\n↻2-∩(×↻2)⊙.,",
     /// Matrix product
     "[7_8_9 10_11_12] [1_2 3_4 5_6]\n⍜⍉⊞(/+×)",
     /// Matrix power (Also works with scalars)
     "4 [1_2 3_4]\n⊙◌⍥(⊞(/+×)⊙⍉,):⊞=.⇡⬚1⊢△,",
     /// Repeat a function and collect intermediate results into an array
     "[⍥(×2.)] 10 1",
     /// Pad an array with 0s
-    "[1_2_3 4_5_6]\n↻⊟.¯1⬚0↙+2△.",
+    "[1_2_3 4_5_6]\n↻∵⋅¯1:⬚0↙+2,⟜△",
```
kaikalii, line 239: What's the rational for this one?

site/src/uiuisms.rs
```diff
@@ -228,15 +228,15 @@ uiuisms!(
     /// Dot product
     "/+× [1 2 3] [4 ¯5 6]",
     /// Cross product
-    "1_2_3 4_5_6\n↻1-∩(×↻1)⊃:⊙∘",
+    "1_2_3 4_5_6\n↻2-∩(×↻2)⊙.,",
```
kaikalii, line 231: I think I prefer `⊃:⊙∘`

kaikalii (changes requested):
Great, thank!
Just a few notes.

site/src/uiuisms.rs
Omnikar: It makes it work for arrays of any rank, whereas the original only worked for rank 2 arrays.

site/src/uiuisms.rs
Omnikar: I think my most preferred way of writing it would be `:,,` since `,,` can be recognized as "dupe the top 2 stack items," but the style checker complains about `:,`.

site/src/uiuisms.rs
kaikalii: If it's just making a list of `¯1`s with the same length as the shape, it should probably use `keep` instead.

Omnikar: Right, I forgot `keep` could do that.

Omnikar: Updated

site/src/uiuisms.rs
kaikalii: You can use `:,,`. I'll fix the lint.

kaikalii (approved):

PR #409, opened 2024-02-21
clarified rule in binding tutorial

GatorHog:
Changed "that is not already used for a Uiua function" to "that is not already used by Uiua" to account for symbols used by Uiua that are not specifically functions, like `[`, `~`, `_`, etc.


kaikalii:
Thanks, but I actually already changed this.

PR #412, opened 2024-02-22
Add bitwise XOR uiuism

bkDJ:
It's probably not the fastest way for 2 scalars, but this way it is rank-polymorphic.


kaikalii:
Can I suggest
```
⍜∩(°⍉⋯)⬚0≠
```

bkDJ:
Of course, that's way better. I've pushed the change.

PR #414, opened 2024-02-23
Use more idiomatic code in `classify` `group` examples

bkDJ:
`fork F G` should be preferred over `F:G.`, no?


kaikalii:
> `fork F G` should be preferred over `F:G.`, no?

Yes. `group` has actually been in the language longer than `fork`; this example hasn't been updated.
Thanks!

PR #441, opened 2024-03-08
Add AB to BAA stack idiom

bkDJ:
Not sure if `⊙.:` would be preferred?


kaikalii:
> Not sure if `⊙.:` would be preferred?

Yes, I think it would

PR #465, opened 2024-03-21
Fix "replace all" uiuism

bkDJ:
* `under keep` can't undo if the destination length differs from the source's
  * I changed `pop` to `reshape shape`
* The uiuism didn't have more than one matching item to showcase the "all" part of the description
  * I added an extra copy of the needle to the haystack


kaikalii:
I think I'm actually going to prefer `⍜▽≡⋅∘=,`.
This is done in 9217683f21db46dcd989f.

bkDJ:
Cool! Without parens is better, but I had forgotten `rows` can resize scalars! 👍

PR #471, opened 2024-04-01
Add "Median"  uiuism

bkDJ:


kaikalii:
Thanks for this!
I think maybe `÷2/+⊏⊟⊃⌊⌈÷2-1⧻.` would be preferable?

bkDJ:
* Your proposal would only work with pre-sorted input, which is why mine selects from a `⍏` first.
* Your proposal raises an error with empty input, but since the other aggregating Uiuisms return `NaN` for that edge case, I wrote this PR to have the same behavior. An array's median should be identical to its arithmetic mean if there are 0, 1, or 2 elements.

What do you think?

kaikalii:
Ah, then perhaps `÷2/+⊏⊟⊃⌊⌈÷2-1⧻⟜(⊏⍏.)`.
I don't think empty should return `NaN`. It should error.

bkDJ:
The latest commit should make it ok 🙂

PR #480, opened 2024-04-10
Update uiuisms

ariymarkowitz:
I think that `∵⋕` is a nicer pattern than `-@0`, it's immediately clear what it does and is shorter. It also verifies that the user actually input a number.

For the Fibonacci number uiuism, it's a small change but I think `:⟜+` is a bit easier to read than `+⟜:`, since there's less nested stack manipulation.

### DIFF
--- a/site/src/uiuisms.rs
+++ b/site/src/uiuisms.rs
@@ -158,9 +158,9 @@ uiuisms!(
     /// Convert a list to a string
     r#"/$"_, _" [10 20 30 40]"#,
     /// Parse a string as a base 2 number
-    r#"°⋯⇌-@0 "110""#,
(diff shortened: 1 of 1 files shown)


kaikalii:
For converting a string of digits to an array of those digits, `-@0` is much faster than `∵⋕`. Character arithmetic is meant to be part of one's Uiua repertoire. 
`⟜:` is an established pattern I want to promote as the primary way of doing the transformation `AB -> ABA`, so `+⟜:` is preferred. In the future, if `by ⊸` is stabilized, there may be cases where `⊸+:` is preferred.
I'll update that Fibonacci example to be what I would consider more idiomatic though.
Thank you for this contribution! I'm going to reject this though.

PR #510, opened 2024-06-06
Fix webkit combinator images

bkDJ:
Webkit doesn't seem to allow CSS into `img` elements. This means Safari and all iPhone browsers have black-on-black letters with serifs in the combinator diagrams.

I use `object` elements to solve this. And to ensure the relative sizes stay the same, I use view boxes in the SVGs. This PR is based on fiddling in my browser, not running the site locally, so consider this un-tested and more of a proposal.

### DIFF
--- a/site/combinators/B.svg
+++ b/site/combinators/B.svg
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="200" height="200">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 200" width="100%" height="100%">
     <style>
         .arg {
             fill: black;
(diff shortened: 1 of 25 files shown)


bkDJ:
It's "[possible](https://chatgpt.com/share/2b6c9ec2-6bf9-4db6-9bca-dcf061bf36ef)" but would involve some ugly workarounds. If someone wants the SVGs, forcing them to "examine the element's HTML" or "look at the source code in this repo" are not great options, but still probably better than polluting CSS or JS for such a rare use case. :/

bkDJ:
Awesome, thanks for merging this, now it looks correct on my tablet:

image

kaikalii:
Ah, the font is still wrong though 😔

PR #512, opened 2024-06-11
Tweak wording on map description

tankorsmash:
This clarifies that `map` takes only two lists. 

I'd have gone a step further and clarify that it puts both the key list and value list back on the stack too, but wasn't sure if that was a welcome change.


tankorsmash:
Accidentally included the now-merged fix from https://github.com/uiua-lang/uiua/pull/511, my mistake!

PR #516, opened 2024-06-27
Allow subscript numbers in identifiers

Omnikar:
Subscripts will format from `__` followed by consecutive digits.
Identifiers are canonicalized during parsing so that `A__1` and `A₁` resolve to the same identifier (`A₁`) even when code is not formatted.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -620,7 +620,7 @@ impl<'a> Formatter<'a> {
                 }
 
                 self.output
-                    .push_str(&crate::parse::canonicalize_exclams(&binding.name.value));
+                    .push_str(&crate::parse::canonicalize_ident(&binding.name.value));
                 self.output
                     .push_str(if binding.public { " ←" } else { " ↚" });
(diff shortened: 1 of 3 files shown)


kaikalii:
Could cananicalization be done in the lexer? I think it makes more sense there.

Omnikar:
That's what I initially thought, but at least from what I've gleaned from the code, the lexer returns identifier tokens as just the span they occupy in the code, so it's not possible for me to change the content of the actual identifier returned by the lexer.

Omnikar:
Also, since there are multiple canonicalization steps now, should they be separated into a complete canonicalization function?

kaikalii:
> Also, since there are multiple canonicalization steps now, should they be separated into a complete canonicalization function like `canonicalize_ident` that just calls `canonicalize_exclams` and `canonicalize_subscripts`?

Yes probably

Omnikar:
Okay I added `canonicalize_ident`.

Issue #526, opened 2024-07-13
Automatically resample decoded audio


yanqui-uxo:
I noticed that when I decoded and attempted to play audio from a .wav file that had a different sample rate than what was used by Uiua, it sounded sped up. I propose that Uiua automatically resample the decoded audio to match the sample rate used. I am willing to make a pull request if this is agreeable.

kaikalii:
I could have sworn I accounted for this.
Do you have somewhere I can get the offending audio file?

kaikalii:
Actually, the reason for this bug makes sense.
Yes, I would accept a PR.

kaikalii:
It's also possible that `un &ae` could return the sample rate.
That way, you could resample it manually with `keep`.
It also would prevent you from mucking up the data by resampling when you don't need to.

PR #528, opened 2024-07-14
Make un&ae automatically resample audio, some changes to soothe clippy::pedantic

yanqui-uxo:
See #526

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3120,6 +3120,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "realfft"
+version = "3.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


kaikalii:
Like I said [here](https://github.com/uiua-lang/uiua/issues/526#issuecomment-2226999842), I don't think this should be done automatically.
I'd prefer `&ae`/`un &ae` to take/return a sample rate.
You could then take the ratio of that sample rate to `&asr`, and use `keep` to resample.
This would also allow the user to encode audio at different sample rates.

kaikalii:
Some new `keep` functionality allows non-integer scalar counts, for scaling arrays.

yanqui-uxo:
I see! Alright, I'll work on your suggestion. Thanks!

kaikalii:
If you do, I think the argument order for `&ae` should be `Format SampleRate SampleData`

PR #530, opened 2024-07-16
Make &ae take audio sample rate, un&ae return audio sample rate

yanqui-uxo:
See discussion at #528

### DIFF
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -440,7 +440,7 @@ sys_op! {
     ///
     /// The first argument is the format, the second is the audio sample rate, and the third is the audio samples.
     ///
-    /// The sample rate must be a 32-bit positive integer.
+    /// The sample rate must be a positive integer.
     ///
     /// The audio samples must be a rank 1 or 2 numeric array.
     ///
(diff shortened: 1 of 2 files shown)


src/sys.rs
```diff
@@ -431,14 +431,16 @@ sys_op! {
...
     ///
     /// See also: [&ae]
-    (1(2), AudioDecode, Audio, "&ad", "audio - decode", Pure),
+    (1(3), AudioDecode, Audio, "&ad", "audio - decode", Pure),
     /// Encode audio into a byte array
     ///
-    /// The first argument is the format, and the second is the audio samples.
+    /// The first argument is the format, the second is the audio sample rate, and the third is the audio samples.
+    ///
+    /// The sample rate must be a 32-bit positive integer.
```
kaikalii, line 443: I don't think the size of the integer should be specified. Nothing else in the docs defines constraints in terms of the size of integers.
I'd prefer just saying "must be a positive integer".
Then, if it is too high to fit in the u32, just error with something like "sample rate is too high".

PR #532, opened 2024-07-31
I added some constants

Migsej:
I felt that these were missing. If there is more please tell me.


kaikalii:
I thought of some of these, but the reason I never added them is that in *all* of these cases, the code to generate them is of less or equal length than the actual name.
```uiua
Digits ← +@0⇡10
AsciiLowercase ← +@a⇡26
AsciiUppercase ← +@A⇡26
AsciiLetters ← ♭⊞+"aA"⇡26
Whitespace ← " \t\n\r"
```

PR #544, opened 2024-09-13
optimize `memberof``range`

Marcos-cat:
tests all work on my machine :+1:


src/algorithm/dyadic/mod.rs
```diff
@@ -1785,3 +1785,59 @@ impl<T: ArrayValue> Array<T> {
...
+
+impl Value {
+    /// Test if a value is an integer in a range
+    pub fn memberof_range(&self, from: Self, env: &Uiua) -> UiuaResult<Self> {
+        fn fallback(a: &Value, b: &Value, env: &Uiua) -> UiuaResult<Value> {
+            let range = a.range(env)?;
+            b.member(&range, env)
+        }
+
+        let Ok(range_bound) = self.as_num(env, "") else {
```
kaikalii, line 1797: I think you should use `.as_nat` here, which returns a `usize` and would eliminate a lot of the checks you do later.

Marcos-cat: I chose to do it this way because I wanted to optimize negatives as well

Marcos-cat: Negatives were included in my original use case. I did have `as_nat` before I got to including them

kaikalii: Ah, okay that makes sense

PR #552, opened 2024-09-25
`gen` upgrade base implementation

amatgil:
Update `gen` to take a shape and a seed and return an array of that shape full of random numbers. There are still some questions to answer, though

- [x] What is the argument order?
- [x] Should the last seed be output as well?

### DIFF
--- a/src/primitive/defs.rs
+++ b/src/primitive/defs.rs
@@ -2639,18 +2639,21 @@ primitive!(
     /// If no value is available, then an error is thrown.
     /// The error can be caught with [try].
     (1, TryRecv, Thread, "tryrecv", Impure),
-    /// Generate a random number between 0 and 1 from a seed, as well as the next seed
+    /// Generate random numbers from a shape with a seed.
     ///
-    /// If you don't care about a seed, you can use [random].
-    ///
-    /// The same seed will always produce the same random number.
-    /// ex: [◌gen gen gen 0]
-    /// ex: [◌⍥gen3 0]
-    /// ex: [◌⍥gen3 1]
+    /// The first argument is the shape, the second argument is the seed. The returned array will have the given shape where each element is in the range [0, 1).
+    /// If you don't care about the seed or shape, [random] may interest you.
+    /// ex: gen 2_3 0
+    /// ex: gen 4 42
+    /// ex: gen 10 42
     ///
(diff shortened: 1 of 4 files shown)


kaikalii:
I think the argument order should be `? Shape Seed`.

It may be simpler to not return the new seed. This could be achieved via `∩gen[],` or the like.

amatgil:
Argument order has been decided on `newseed result ? shape seed`

It should be ready to merge

src/primitive/defs.rs
```diff
@@ -2639,18 +2639,21 @@ primitive!(
     /// If no value is available, then an error is thrown.
     /// The error can be caught with [try].
     (1, TryRecv, Thread, "tryrecv", Impure),
-    /// Generate a random number between 0 and 1 from a seed, as well as the next seed
+    /// Generate random numbers from a shape with a seed.
```
kaikalii, line 2642: Maybe "Generate an array of random numbers with a seed"

src/primitive/defs.rs
```diff
@@ -2639,18 +2639,21 @@ primitive!(
...
     ///
-    /// If you don't care about a seed, you can use [random].
-    ///
-    /// The same seed will always produce the same random number.
-    /// ex: [◌gen gen gen 0]
-    /// ex: [◌⍥gen3 0]
-    /// ex: [◌⍥gen3 1]
+    /// The first argument is the shape, the second argument is the seed. The returned array will have the given shape where each element is in the range [0, 1).
+    /// If you don't care about the seed or shape, [random] may interest you.
```
kaikalii, line 2645: "... you can use [random] instead."

src/primitive/defs.rs
```diff
@@ -2639,18 +2639,21 @@ primitive!(
...
+    /// ex: gen 10 42
     ///
     /// Use [multiply] and [floor] to generate a random integer in a range.
-    /// ex: ⌊*10[◌⍥gen5 0]
-    (1(2), Gen, Misc, "gen"),
+    /// ex: ⌊×10 gen 2_3 0
+    /// ex: ⌊×10 gen 10 42
+    //
+    /// For non-determinism, [random] can be used as a seed.
+    /// ex: gen 5 ⚂
```
kaikalii, line 2655: Maybe a 2D shape here?

PR #555, opened 2024-09-28
fix issue where prim links in the changelog require a full name

Marcos-cat:


kaikalii:
Should probably just use `Primitive::from_format_name`

Issue #38, opened 2023-09-30
Sixel encoding for images and output for terminals


TeddyDD:
Just an idea, sixel format could be used to display images in terminal.

kaikalii:
I support this.
I have other things higher in my todo, but if anyone wants to implement this, know that you cannot use any of the existing Rust crates that depend on libsixel, as they require it as a local dependency - something I do not want to requre.
I suggest looking at [icy_sixel](https://crates.io/crates/icy_sixel)

exxjob:
Just to note: terminal emulators have different ideas of what's worth supporting, so compatibility can be a pain. There's ReGIS, Sixel, Kitty, iTerm2, and finally mosaics and ueberzug which are entirely different things. You might find Chafa and NotCurses most flexible, both seem to have Rust wrappers, but I can't vouch for those. https://www.arewesixelyet.com

kaikalii:
It would be part of &ims.

JoaoFelipe3:
This should get closed, since #557 just got merged.

TeddyDD:
Can we get other method of enabling it? I'd rather not mess with TERM variable. Perhaps env like `UIUA_ENABLE_SIXEL` or something?

BaerLKR:
> Can we get other method of enabling it? I'd rather not mess with TERM variable. Perhaps env like `UIUA_ENABLE_SIXEL` or something?

that would easily be possible

TeddyDD:
Since Uiua supports Sixel images now, I think this can be closed. Thanks @BaerLKR!

PR #557, opened 2024-09-28
added sixel encoding

BaerLKR:
This is in regards to this https://github.com/uiua-lang/uiua/issues/38 issue, as suggested i used icy_sixel

It encodes to sixel if it finds `sixel` in the `TERM` environment variable. That sadly isn't always the case but I couldn't find another way to detect if a terminal supports sixel. And I wanted to prevent false-positive sixel output because if you don't know what it is this string of random chars is rather confusing. So there will be quite a few false-negatives but there is still the half block fallback from `viuer`

I hope this is what y'all had in mind and I am looking forward to your criticism

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1719,6 +1719,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "icy_sixel"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "86858ae800284d596cfdefcb0ad435c3493c12f35367431bbe9b2b3858c1155b"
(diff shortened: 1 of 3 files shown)


kaikalii:
I think you need to use [`option_env`](https://doc.rust-lang.org/std/macro.option_env.html)

BaerLKR:
yeah i just noticed too
will fix in just a second

BaerLKR:
> I think you need to use [`option_env!`](https://doc.rust-lang.org/std/macro.option_env.html)

I think I need to use `std::env::var` because it is a runtime check

PR #559, opened 2024-09-30
Extend parse's implementation to accept both formats

amatgil:


kaikalii:
Can you please add tests to `tests/units.rs`?

amatgil:
Certainly
(Though a type of each case is in the documentation and gets run as a test already)

tests/units.ua
```diff
@@ -211,6 +211,8 @@
 ⍤⤙≍ [2 10 5] ⬚5⋕{"2" "10" "dog"}
 ⍤⤙≍ [ℂ 3.4e6 1.2 ℂ π ∞] ⋕ {"1.2r3.4e6i" "infrpii"}
 ⍤⟜≍ ⋕°⋕. [ℂ 3.4e6 1.2 ℂ π ∞]
+⍤⤙≍ [ℂ4.5 `12.3 ℂ4.5 12.3 ℂ`4.5 12.3 ℂ`4.5 12.3 ℂ4.5 12.3] ◇⋕ {"-12.3r4.5i" "12.3r4.5i" "12.3-4.5i" "12.3r-4.5i" "12.3r4.5i"}
```
kaikalii, line 214: Why `◇ content`?

Can you add some tests to verify that integers also work?
And `pi`, `inf` etc.

amatgil:
I might've added too many tests, I can't tell what's too much and what's too little

PR #560, opened 2024-10-01
Implement around as a complement to over

amatgil:
It's missing a properly worded documentation and to decide if it would be a good addition to the language


kaikalii:
I'll merge this if you add `around` to `fn is_experimental`

amatgil:
I think that's everything

PR #561, opened 2024-10-01
adding a sixel enabeling env var 

BaerLKR:
sorry for making yet another PR regarding the same topic and including only a small change

but here i addressed some points made in https://github.com/uiua-lang/uiua/issues/38 issue

I added the option to force-enable sixel via an environment variable and documented it in the docs for `&ims` in I way i thought was fitting

hope this helped

### DIFF
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -370,7 +370,7 @@ sys_op! {
     /// How the image is shown depends on the system backend.
     ///
     /// In the default backend, the image is shown in the terminal. Here you can make it use sixel
-    /// by setting the `UIUA_ENABLE_SIXEL` environment variable to `1`. Else it will try to use the
+    /// by setting the `UIUA_ENABLE_SIXEL` environment variable to `1`. Otherwise it will try to use the
     /// `kitty` or `iTerm` image protocols and fall back to half-block image printing.
     /// On the web, the image is shown in the output area.
     ///
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -369,7 +369,9 @@ sys_op! {
     ///
     /// How the image is shown depends on the system backend.
     ///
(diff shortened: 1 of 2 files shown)


src/sys.rs
```diff
@@ -369,7 +369,9 @@ sys_op! {
     ///
     /// How the image is shown depends on the system backend.
     ///
-    /// In the default backend, the image is shown in the terminal.
+    /// In the default backend, the image is shown in the terminal. Here you can make it use sixel
+    /// by setting the `UIUA_ENABLE_SIXEL` environment variable to `1`. Else it will try to use the
```
kaikalii, line 373: Should be "Otherwise it will..."

src/sys_native.rs
```diff
@@ -587,6 +587,7 @@ impl SysBackend for NativeSys {
         if std::env::var("TERM")
             .unwrap_or("".to_owned())
             .contains("sixel")
+            || std::env::var("UIUA_ENABLE_SIXEL") == Ok(String::from("1"))
```
kaikalii, line 590: Should be `std::env::var("UIUA_ENABLE_SIXEL").is_ok_and(|s| s == "1")`

src/sys.rs
```diff
@@ -369,7 +369,9 @@ sys_op! {
     ///
     /// How the image is shown depends on the system backend.
     ///
-    /// In the default backend, the image is shown in the terminal.
+    /// In the default backend, the image is shown in the terminal. Here you can make it use sixel
+    /// by setting the `UIUA_ENABLE_SIXEL` environment variable to `1`. Otherwise it will try to use the
+    /// `kitty` or `iTerm` image protocols and fall back to half-block image printing.
```
kaikalii, line 374: Paragraphs should be on a single line.

PR #562, opened 2024-10-02
filled csv

Marcos-cat:


tests/units.ua
```diff
@@ -315,8 +315,7 @@ F ← ⨬(¯|+1|×10|∞)
 ⍤⤙≍ [[[1_2_3 5_6_7]] [[9_10_11 13_14_15]]] ◫¤2_3 +1°△4_4
 ⍤⤙≍ [
   [1_2_3 5_6_7]_[4_0_0 8_0_0]
-  [9_10_11 13_14_15]_[12_0_0 16_0_0]
-] ⬚0◫¤2_3 +1°△4_4
+  [9_10_11 13_14_15]_[12_0_0 16_0_0]] ⬚0◫¤2_3 +1°△4_4
```
kaikalii, line 318: You seem to be using an old version of the formatter. Please remove this change.

tests/units.ua
```diff
@@ -709,7 +708,9 @@ B ← |1 (⨬(+⊃(B-1|B-2)|1)<2.)
 ⍤⤙≍ ⋕ ⍜csv∘ . ↯3_4⇡12
 ⍤⤙≍ ∵⍣⋕∘ ⍜csv∘ . [{"Foo" "Bar"} [1 2] [3 4] [5 6]]
 ⍤⤙≍ [{"1" "2" ""} {"3" "" ""} {"4" "5" "6"}] °csv "1,2\n3\n4,5,6\n"
+⍤⤙≍ [{"1" "2" ""} {"3" "" ""} {"4" "5" "6"}] ⬚@;°csv "1;2\n3\n4;5;6\n"
```
kaikalii, line 711: I don't think this behavior should be for `fill(..) un csv`.
It should be for `un fill(..) csv`.
I'll use this PR as a starting point to enabled that.

Marcos-cat: I don't think I'm knowledgable enough to implement this in any measure.

PR #567, opened 2024-10-03
refactor(flake): clean up the flake

thecaralice:
1. Add a cargo feature named `system` which causes dependencies to use system versions of static libraries instead of compiling their own, which is particularly important to Nix (previously the Nix derivation would patch `Cargo.toml` to enable the corresponding `libffi` feature).
2. Separate the package derivation from the flake itself.
3. Properly clean source for the derivation, so changing files that are not used for building would not make Nix think everything needs to be rebuilt.
4. Remove `rust-overlay` dependency. It is great, but unnecessary here, as Uiua does not need bleeding-edge toolchain versions or specific components and targets. It would make sense to use it if the website was built with Nix as well.
5. Use `flake-parts` instead of `flake-utils` as it doesn't have design flaws `flake-utils` has and fits Nix better.
6. Remove input overrides from `crane` as `crane` doesn't have any inputs anymore.
7. Generally clean up Nix code, as there was some suspicious things like calling an ad hoc constructed function (`(path: _type: ...) path type`).
8. Build dependencies separately from the main derivation to improve incremental compilation.

This also revealed a couple of suspicious things like the tests depending on the website favicon and one of the tests actually trying to write some file, so I decided to disable it when building with Nix (`format::generate_format_cfg_docs`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,6 +149,10 @@ tls = ["httparse", "rustls", "webpki-roots", "rustls-pemfile"]
 web = ["wasm-bindgen", "js-sys", "web-sys"]
 webcam = ["image", "uiua-nokhwa"]
 xlsx = ["calamine", "simple_excel_writer"]
+# Use system static libraries instead of building them
+system = [
+  "libffi?/system"
+]
 
 [[bin]]
 name = "uiua"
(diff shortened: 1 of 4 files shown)


kaikalii:
I don't know nix. Can you please explain what this changes?

thecaralice:
Sorry, forgot to write the description. Updated.

PR #568, opened 2024-10-04
update the tour

Marcos-cat:


kaikalii:
I'm not sure modifiers should be added to some of this code.

Although it raises a question about whether things like `on`/`by` (and maybe `below`) should be taught earlier in the tour/tutorial.

Marcos-cat:
My opinion was that, Because the code got more complicated as it went down, `by` was not the most complicated thing they would encounter so was fine to include.

Marcos-cat:
I don’t think they should be taught, but showing them in later examples seemed harmless

Marcos-cat:
abandoning this until we talk about it further

PR #574, opened 2024-10-09
add `&clip`

Marcos-cat:


site/src/editor/mod.rs
```diff
@@ -54,7 +54,7 @@ pub fn Editor<'a>(
 ) -> impl IntoView {
     let no_run = no_run
         || mode == EditorMode::Pad && !get_autorun()
-        || ["&sl", "&httpsw", "send", "recv", "&ffi", "&clset"]
```
kaikalii, line 57: All these removed references to `&clset` in should be replaced with `°&clip`.
Tests and things should be able to set the clipboard.

Marcos-cat: should it just disallow `&clip` since someone could reasonably write `°(&clip)`

kaikalii: Yes I think so

PR #580, opened 2024-10-15
Fix 0.12.3 on Nix

prsteele:
This updates the `0.12.3` release to build on Nix. In particular, it will be possible to run

```
nix run github:uiua-lang/uiua/0.12-fix -- repl
```

to enter a Uiua REPL.


kaikalii:
Sorry, I prematurely merged this then had to revert it.
Dependencies cannot be `git`, they must be based on published version, otherwise the crate cannot be published on crates.io.

prsteele:
Ah, I see. That's unfortunate; the good news is that, for whatever reason, the code on `main` seems perfectly happy building without these changes.

PR #584, opened 2024-10-17
add anti math style warnings

Marcos-cat:


kaikalii:
You may need to remove the examples that produce this warning.

But also, I wonder if these warning are necessary?

Marcos-cat:
I guess probably not in Uiua's current state. I just hope for a Uiua with suggestions like clippy one day.

PR #588, opened 2024-10-19
Remove deal & optimize firunsort

Marcos-cat:


src/optimize.rs
```diff
@@ -57,6 +57,11 @@ pub(crate) fn optimize_instrs_mut(
             instrs.pop();
             instrs.push(Instr::ImplPrim(MemberOfRange, span))
         }
+        // First UnSort = random row
+        ([.., Instr::ImplPrim(UnSort, _)], Instr::Prim(First, span)) => {
```
kaikalii, line 61: Should do `last` as well.

kaikalii: `Instr::Prim(First | Last, span)`

Omnikar: I think this would also require changing the error message at algorithm/dyadic/mod.rs:1943

kaikalii: I'm not opposed to the error just saying something like "Can't pick random row from empty array."

Marcos-cat: Oh my lord, you guys are replaying my exact conversation with myself from two minutes before I PRd this

Marcos-cat: (I’ll add it)

PR #594, opened 2024-10-20
Optimize `∈☇1⇡` for multidimensional ranges

Omnikar:
This optimizes patterns like `∈☇1⇡ [50 50]` to check if the rank 1 sub-rows of an array are completely within the range bounds.

Marcos mentioned that it might be odd to have this as a separate `ImplPrim`, but I feel like the implementation is different enough from the normal `∈⇡` optimization to warrant it.


Omnikar:
TODO: Fix how this works on scalar ranges
`∈☇1⇡ 10 ◫¤¤¯2⇡20` currently evaluates to `[1 0]` with the normal code, but `[0 0]` with the optimization.

PR #598, opened 2024-10-28
chore: bump `viuer` to `0.9.1`

magistau:
In particular this includes support for Ghostty.
Depends on uiua-lang/nokhwa#1.


dilr:
This conflicts now because I made a now accepted PR updating `viuer` to 0.9.1 as a side effect of updating `image` as a side effect of updating `nukhwa`. Note that my PR changed `uiua` from depending on `uiua-nukhwa` to relying on `nukhwa` itself. So I think this PR can be closed. The other PR probably doesn't conflict, but I don't know if kaikalii will bother to maintain the fork now that uiua can use mainline `nukhwa`.

PR #599, opened 2024-10-29
Make more things public

ekgame:
This change exposes a couple of struct fields for the public API. These are the fields I currently found useful for inspecting parsed code structure to reinterpret the code in a format that could be consumed outside Rust (with the goal to create a documentation generator). 

The two fields are:
- `Assembly#inputs` - The library may load more than one file and it's useful to know what files were used.
- `Function#signature` - When inspecting a function it's useful to know it's inferred signature.


kaikalii:
Function signature is private so that it cannot be changed. Use the getter.

kaikalii:
Actually maybe it's fine.

PR #603, opened 2024-11-15
document the printing functions

Marcos-cat:


src/sys.rs
```diff
@@ -161,10 +161,30 @@ impl SysOpClass {
 
 sys_op! {
     /// Print a nicely formatted representation of a value to stdout
+    ///
+    /// [&s] will print the value the same way it would appear at the end of a program, or from [?]
```
kaikalii, line 165: Lines after the first one should end in a period.

PR #621, opened 2024-12-06
Add "Repeat until error is thrown" idiom

lemonlime4:


kaikalii:
I'm curious, do you have a concrete use case for this?

It should probably be [`⍢∘⍣⊓1F0`](https://uiua.org/pad?src=0_14_0-dev_7__RiDihpAgJnDCsOKKggrijaLiiJjijaPiipMxRjAg4oehNAo=).

lemonlime4:
Ran into this while doing advent of code 2024 day 6

PR #622, opened 2024-12-08
Add support for fill in (anti)base

bluebear94:
For array bases, the fill value is used when the base itself is exhausted. In other words, `⬚r base B` is equivalent to `⊂⊃(⍜⇌↘₁|base r⊣) base ⊂B∞`. `⌝base` gets a similar change for consistency.


kaikalii:
This is great, thank you! One thing, would you mind adding an example or two to `src/primitive/defs.rs`?

Issue #628, opened 2024-12-15
Shebang scripts parse arguments after the script path as regular args


nilium:
I noticed in 0.14.0-rc.1 that scripts where I'd pass a test input into a script didn't run as usual (a side effect of how I structured Advent of Code stuff). This can be pretty easily checked with a script like:

```uiua
#!/usr/bin/env uiua
# thing.ua
&p &args
```

Running it as `./thing.ua test` doesn't run the script itself and instead runs the test subcommand (this is a little easier to see with `lsp` instead, but I first noticed it with `test`).

Works under 0.13.0, so can probably narrow it down to changes around arg parsing since. I'll probably take a go at fixing it but wanted to get this written down for now since it's getting late here. I think maybe something related to 0f38f26003a56a5390181f3f8fa42b990c7839e2 since it changed arg handling after #585 / aee40c5b6c92d8c1824590406c2f18a5e31b42dd, but that's from a very, very quick scan.

kaikalii:
I saw the PR, but I'm not sure it's the best way.
I'm not on my Linux machine right now, but I'll test solutions to this tomorrow when I am!

PR #630, opened 2024-12-16
Move shebang handling up before Clap arg parsing (#628)

nilium:
With the change from try_parse to parse and using app.file to get a file passed before a command, it's possible for clap to parse `uiua file.ua test` and populate both app.file and app.command. This can break shebang scripts if the script accepts any argument that clap can parse (subcommands but also possibly other arguments). Moving app.file checks up before app.command checks unfortunately don't fix this, since clap will parse e.g., `--help` and print usage output and exit before Uiua can handle args itself.

Fix this by going back to something similar to earlier arg parsing, but check the first argument for something path-like. I don't want to inadvertently break command handling if e.g., a file named `help` is in the current directory, so a file matching the first argument doesn't feel like it's enough to do this. Instead, add something that specifically looks for a first argument that has a path separator and then short circuit arg parsing for that. This handles a few main cases I can run locally:

  * Relative `path/to/script args...` -- this presents the first argument as `path/to/script`, so it necessarily contains a path separator. This includes the usual `./script args...` invocations.

  * Running `PATH=. script args` -- same outcome as the previous case, runs the script as `./script`, so a path separator still occurs.

  * Running `PATH=/abs/dir-with-script` -- the script argument becomes `/abs/dir-with-script/script`, so again contains a path separator.

Not sure how Windows handles this sort of thing, I'd guess WSL and co. follow the same rules for shebangs as anything else, so the above should all still apply. Not sure about the rest, less sure about Uiua support for Windows to begin with too.

Remove the `file` field from `App` since it's unused with this change.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -112,6 +112,29 @@ fn main() {
         print_stack(&rt.take_stack(), true);
         return;
     }
+
+    // Check for a path-like argument prior to any other args. If found (must contain a path
+    // separator; run as ./foo or with an absolute path by being found in PATH), run without
+    // argument parsing or formatting.
+    let mut args = std::env::args().skip(1);
+    if let Some(path) = args
+        .next()
+        .filter(|path| path.contains(std::path::is_separator))
+        .map(PathBuf::from)
+    {
+        run(
+            path.as_ref(),
+            args.collect(),
+            false,
+            None,
+            Some(RunMode::Normal),
+            None,
+            false,
+        );
+        return;
+    }
+    drop(args);
+
     let app = App::parse();
     match app.command {
         Some(Comm::Init) => {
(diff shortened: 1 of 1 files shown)


nilium:
Not sure if this would be preferred to going back to try_parse, but I think the `file` field on the `App` struct probably isn't work-able with how Clap works. Not sure there's a way to make this work _better_ with Clap either, but seemed easier to check the first arg for something path-like than to do more involved things through Clap. No worries about closing this if there's a better route though.

kaikalii:
In investigating alternatives to this implementation, I settled on something very similar.
Fixed in e8b3be79318bf7753b7

nilium:
Nice, thanks for the help!

PR #631, opened 2024-12-16
added subscript docs to sqrt page

BaerLKR:
I noticed it was missing a notice on the sqrt doc page that subscripts are available

### DIFF
--- a/src/primitive/defs.rs
+++ b/src/primitive/defs.rs
@@ -655,6 +655,10 @@ primitive!(
     /// You can only take the square root of a negative number if it is complex.
     /// ex: √  ¯4
     ///   : √ℂ0¯4
+    ///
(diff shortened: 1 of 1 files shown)


BaerLKR:
Would it be possible/wanted to automatically add this to the docs when there is an entry to that primitve in the subscripts doc page? That would make it impossible to change docs in one place and forget to change it in another

kaikalii:
> Would it be possible/wanted to automatically add this to the docs when there is an entry to that primitve in the subscripts doc page? That would make it impossible to change docs in one place and forget to change it in another

Hmm, maybe I could add support for a primitive's docs to specifically have `un`/`under`/`fill`/subscript sections that would automatically get added at the appropriate places.

PR #637, opened 2024-12-21
Added a function to get normalized keys from a map

ekgame:
This PR adds a function to return map keys as a value where each row is in the same order as the array values.

I am making a Uiua runtime library for JavaScript (https://github.com/ekgame/uiua-js) and I want to convert from Rust API's Value struct to a JSON object with all the equivalent information. The one thing stopping me is that the keys for maps are all jumbled up internally and include dummy values. I need this function as a utility to get the keys in the same order as the array values so I could properly represent a Uiua value in JavaScript.


ekgame:
Looks like I missed the `MapKeys::normalized` function that effectively does the same thing, closing.

PR #642, opened 2024-12-31
experimental negate subscripts as roots of unity

Marcos-cat:


Marcos-cat:
Experimental so it can die if it is unliked

PR #649, opened 2025-01-20
Use bytemuck to replace most instances of mem::transmute.

dilr:
Note that I am using bytemuck to create float constants. This works only if u64 and f64 have the same endianness. They do on all platforms I know of, so that shouldn't be a problem, but it would be better to use f64::from_bits for this. Unfortunately, that function isn't stable until rust 1.83, which is below the current MSRV for this project according to Cargo.toml.


kaikalii:
Thanks!

PR #653, opened 2025-02-02
fix(nix): use makeBinaryWrapper

thecaralice:
It is recommended to use this, IBNLT because macOS expects shebang targets to be binaries. See [docs].

[docs]: https://nixos.org/manual/nixpkgs/unstable/#fun-makeWrapper

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -1,7 +1,7 @@
 {
   craneLib,
   libPath,
-  makeWrapper,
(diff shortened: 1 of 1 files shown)


TomaSajt:
I never actually read this part of the manual, so I never knew about this limitation.
I'll try to get this fixed downstream in nixpkgs as well.

PR #654, opened 2025-02-05
Download all .ua files from a GitHub repo in the pad

ekgame:
After some investigation, I found a way to introspect a GitHub repo to list all the files in a repository via the GitHub API. Turns out that the endpoint doesn't require an auth token and the GitHub API is CORS friendly - just what we need to allow multi-file git modules in the pad.

List of changes:
- This PR implements the logic to download a list of files from a GitHub repo, finds the `.ua` files and downloads all of them preserving the repo structure.
- Because this may create a bunch of files with long paths in the virtual file system - the pad also hides all files whose the path begins with "uiua-modules/".

*Note*: This is not a universal solution to all Git repo hosting providers, it only supports GitHub and any other providers would need to be implemented manually assuming they also have similarly capable API. While that's not great, I've only seen the community use GitHub links and this change would allow sharing code that depends on more complex modules.

To test the changes, I have set up a repository with multiple `.ua` files to see how the files are loaded and validate that relative imports within a module work: https://github.com/ekgame/test-multi-file-uiua-repo.

To test the changes, simply run:
```
~ "git: github.com/ekgame/test-multi-file-uiua-repo"
```
And you should see this output:
```
Printing from foo.ua
Printing from folder/bar.ua
Printing from lib.ua
```


ekgame:
I believe the change to the TREE_URL formatting broke the url, testing the same code on the website displays a 404 error ([pad link](https://www.uiua.org/pad?src=0_15_0-dev_1__fiAiZ2l0OiBnaXRodWIuY29tL2VrZ2FtZS90ZXN0LW11bHRpLWZpbGUtdWl1YS1yZXBvIgo=))

kaikalii:
Oh ya lol my bad

PR #658, opened 2025-02-22
fix markdown parser of primitive's documentation

halfminami:
I noticed some markdown on docs is broken. This fixes it.
If a line contains strong text, the format of the line is not converted into HTML. Namely [each](https://www.uiua.org/docs/each) and [bits](https://www.uiua.org/docs/bits). You can see square brackets of function names are not turned into links although other lines are.
(Texts shown on hovering those functions using the language server in my text editor also have the same issue.)

In the code, the case of the changed line should indicate the end of the strong text. So it doesn't need to check if the line has another `**`, which I removed.
(I didn't know how to test things, but I locally built and checked the website and the language server to ensure that they are working correctly on those functions, at least.)


kaikalii:
Awesome, thank you!

Issue #663, opened 2025-02-23
Proposal: make "requirement" parameter in Value.as_* methods Option<&'static str>


yanqui-uxo:
I've noticed that these methods take a string as a requirement message, but for most of them, if the string is empty, it replaces it with a default message. I don't think this behavior is very clear and would be better represented with an `Option` as the parameter. I also believe the ability to use a default message should be added to all of these methods.

yanqui-uxo:
Would you be open to a pull request changing the API to use `Option`?

kaikalii:
Yes

PR #667, opened 2025-02-27
Make pub Value.as_* methods take an Option as a requirement

yanqui-uxo:
See #663


kaikalii:
Looking at this, I think I'd prefer `requirement: impl Into<Option<&'static str>>` so you don't have to wrap everything in `Some`.

PR #673, opened 2025-03-09
Update nixpkgs

ursi:
Rust version was too old to build


ursi:
wow, merged so fast I went back to see if github cli had thrown an error (cuz I didn't see the PR) :joy:

PR #675, opened 2025-03-22
add tuples infinity

Marcos-cat:


kaikalii:
Can you add it to the `tuples` docs?

Marcos-cat:
Sure

PR #684, opened 2025-04-04
stabilize negate subscripts

Marcos-cat:


src/primitive/defs.rs
```diff
@@ -199,6 +199,9 @@ primitive!(
     /// ex: ¯ "Hello, World!"
     /// Use this with [absolute value] to lowercase a string.
     /// ex: ¯⌵ "Hello, World!"
+    ///
+    /// Subscripted [negate] will rotate a number one nth of a rotation in the complex plane.
+    /// ex: ▽⟜≡▽10°⊚≡⊟°ℂ-⊸/↧ . ⁅ [◌⍥₁₀⊸¯₁₀] 10
```
kaikalii, line 204: I think this example should be much simpler.
Maybe stuff like
```
¯₂ 5
¯₄ 5
⁅₃ ¯₃ 5
[⍥₃⊸¯₄ ℂ1 1]
```

PR #685, opened 2025-04-05
add swap alias for flip

haydenflinner:
I was tripped up a few times on my first morning with uiua by typing "swap" rather than "flip" 😄


kaikalii:
One issue with this is that it could also be an alias for `backward`.

haydenflinner:
Interesting, TIL of `backward`! `swap` doesn't have a link to `backward`, but `backward` seems to be a different "time of applying" `swap`.  I'll add that backlink if I get a chance.

I've also noticed that searching "print" doesn't bring up the "pf" command for "print and flush". 

Perhaps the search bar including descriptions would fix both "searching for swap from memory" and "searching for print out of familiarity"! I may take a look at that too.

Do you have any plans for a more interactive/visual debugger written anywhere?

kaikalii:
The "time of applying" is the same. `backward` is just a modifier while `flip` is a function.

I don't have plans for visual debugging. That's what stuff like `stack ?` or output comments are for.

kaikalii:
`flip` is generally discouraged now, so I'll close this.

PR #688, opened 2025-04-08
Add small recap the Fix section of the Advanced Array Tutorial

amatgil:
I used labels for one of the examples, which might not be acceptable. If not, they can be removed directly

### DIFF
--- a/site/src/tutorial.rs
+++ b/site/src/tutorial.rs
@@ -1357,24 +1357,37 @@ fn TutorialMoreArray() -> impl IntoView {
         <Editor example="≡⊂ 1_2_3 4_5_6"/>
         <p>"Usually, the arrays must have the same number of rows."</p>
         <Editor example="≡⊂ 1_2_3 4_5"/> // Should fail
-        <p>"However, there is an exception for arrays that have exactly one row. In this case, that row will be repeated for each row of the other array(s)."</p>
+        <p>"However, there is an exception for scalars and for arrays that have exactly one row. In this case, that scalar/row will be repeated for each row of the other array(s)."</p>
         <Editor example="≡⊂ 1_2_3 4"/>
-        <Editor example="≡⊂ 1 2_3_4"/>
+        <Editor example="≡⊂ 4 1_2_3"/>
         <Editor example="≡(⊂⊂) 1 2_3_4 5"/>
-        <p>"Notice that the second argument here is a 2D array with 1 row of 2 elements. It will be repeated just like the scalars above."</p>
+        <p>"Notice that the second argument of this next example is a 2D array with 1 row of 2 elements. It will be repeated just like the scalars above."</p>
         <Editor example="≡⊟ [1_2 3_4 5_6] [¯1_0]"/>
-        <p>"If we want to combine each row of one array with copies of another, we can turn one of the arrays into a single row array with "<Prim prim=Fix/>". "<Prim prim=Fix/>" adds a 1 to the front of the shape of an array."</p>
-        <Editor example="¤.1_2_3"/>
+        <p>"If we want to combine each row of one array with copies of another, we can turn one of the arrays into a single row array with "<Prim prim=Fix/>". "<Prim prim=Fix/>" adds a 1 to the front of the shape of an array, and is equivalent to wrapping it in a pair of square brackets. Both of these methods create an array that only contains the singular row."</p>
+            <Editor example="$WithoutFixing 1_2_3
+$Fixed ¤1_2_3
+$Bracketed [1_2_3]"/>
         <p>"Here, we "<Prim prim=Fix/>" "<code>"1_2_3"</code>" so that it is reused for each row of "<code>"4_5_6"</code>"."</p>
         <Editor example="≡⊂ ¤ 1_2_3 4_5_6"/>
         <p>"If we have a bunch of arrays and want to choose which ones are fixed and which are not, we can use planet notation."</p>
         <Editor example="≡⊂ ⊙¤ 1_2_3 4_5_6"/>
-        <Editor example="≡(⊂⊂⊂) ⊓⊓⊓∘¤¤∘ 1_2_3 4_5_6 7_8_9 10_11_12"/>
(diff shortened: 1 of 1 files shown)


site/src/tutorial.rs
```diff
@@ -1357,24 +1357,37 @@ fn TutorialMoreArray() -> impl IntoView {
...
         <Editor example="≡(⊂⊂) 1 2_3_4 5"/>
-        <p>"Notice that the second argument here is a 2D array with 1 row of 2 elements. It will be repeated just like the scalars above."</p>
+        <p>"Notice that the second argument of this next example is a 2D array with 1 row of 2 elements. It will be repeated just like the scalars above."</p>
         <Editor example="≡⊟ [1_2 3_4 5_6] [¯1_0]"/>
-        <p>"If we want to combine each row of one array with copies of another, we can turn one of the arrays into a single row array with "<Prim prim=Fix/>". "<Prim prim=Fix/>" adds a 1 to the front of the shape of an array."</p>
-        <Editor example="¤.1_2_3"/>
+        <p>"If we want to combine each row of one array with copies of another, we can turn one of the arrays into a single row array with "<Prim prim=Fix/>". "<Prim prim=Fix/>" adds a 1 to the front of the shape of an array, and is equivalent to wrapping it in a pair of square brackets. Both of these methods create an array that only contains the singular row."</p>
+            <Editor example="$WithoutFixing 1_2_3
+$Fixed ¤1_2_3
+$Bracketed [1_2_3]"/>
```
kaikalii, line 1369: I don't think we should use labels here, as they aren't introduced yet.

site/src/tutorial.rs
```diff
@@ -1357,24 +1357,35 @@ fn TutorialMoreArray() -> impl IntoView {
         <Editor example="≡⊂ 1_2_3 4_5_6"/>
         <p>"Usually, the arrays must have the same number of rows."</p>
         <Editor example="≡⊂ 1_2_3 4_5"/> // Should fail
-        <p>"However, there is an exception for arrays that have exactly one row. In this case, that row will be repeated for each row of the other array(s)."</p>
+        <p>"However, there is an exception for scalars and for arrays that have exactly one row. In this case, that scalar/row will be repeated for each row of the other array(s)."</p>
         <Editor example="≡⊂ 1_2_3 4"/>
-        <Editor example="≡⊂ 1 2_3_4"/>
+        <Editor example="≡⊂ 4 1_2_3"/>
```
kaikalii, line 1362: No reason to change this.

site/src/tutorial.rs
```diff
@@ -1357,24 +1357,35 @@ fn TutorialMoreArray() -> impl IntoView {
...
         <Editor example="≡(⊂⊂) 1 2_3_4 5"/>
-        <p>"Notice that the second argument here is a 2D array with 1 row of 2 elements. It will be repeated just like the scalars above."</p>
+        <p>"Notice that the second argument of this next example is a 2D array with 1 row of 2 elements. It will be repeated just like the scalars above."</p>
         <Editor example="≡⊟ [1_2 3_4 5_6] [¯1_0]"/>
-        <p>"If we want to combine each row of one array with copies of another, we can turn one of the arrays into a single row array with "<Prim prim=Fix/>". "<Prim prim=Fix/>" adds a 1 to the front of the shape of an array."</p>
-        <Editor example="¤.1_2_3"/>
+        <p>"If we want to combine each row of one array with copies of another, we can turn one of the arrays into a single row array with "<Prim prim=Fix/>". "<Prim prim=Fix/>" adds a 1 to the front of the shape of an array, and is equivalent to wrapping it in a pair of square brackets. Both of these methods create an array that only contains the singular row."</p>
+        <Editor example="¤1_2_3 [1_2_3] # Equivalent"/>
```
kaikalii, line 1367: Should have newlines between them I think.

site/src/tutorial.rs
```diff
@@ -1357,24 +1357,35 @@ fn TutorialMoreArray() -> impl IntoView {
...
         <Editor example="≡⊂ ¤ 1_2_3 4_5_6"/>
         <p>"If we have a bunch of arrays and want to choose which ones are fixed and which are not, we can use planet notation."</p>
         <Editor example="≡⊂ ⊙¤ 1_2_3 4_5_6"/>
-        <Editor example="≡(⊂⊂⊂) ⊓⊓⊓∘¤¤∘ 1_2_3 4_5_6 7_8_9 10_11_12"/>
         <Editor example="≡(⊂⊂⊂) ⊙∩¤     1_2_3 4_5_6 7_8_9 10_11_12"/>
         <p><Prim prim=Fix/>" also works without "<Prim prim=Rows/>" with pervasive dyadic functions."</p>
         <Editor example="-  [1 2 3]  [4 5 6]\n- ¤[1 2 3]  [4 5 6]\n-  [1 2 3] ¤[4 5 6]"/>
         <Editor example="-  1_3 [3_4 5_6 7_8]"/> // Should fail
         <Editor example="- ¤1_3 [3_4 5_6 7_8]"/>
+        <h3>"Recap"</h3>
```
kaikalii, line 1377: What's the purpose of a recap? No other tutorial section has one. If more explanation of some of these concepts is required, it should go above.

site/src/tutorial.rs
amatgil: I think it illustrates the point way better if the scalar keeps being a scalar, no?

site/src/tutorial.rs
amatgil: I remember reading through this section and feeling like it was  "neat, but it'll probably be very uncommon. Why does it need its own glyph, even?"
Someone asked in the discord for an explanation of what fixing actually did, because (I presume) they got the same feeling
I think a very short paragraph tldr-ing what it does would help digest the explanation faster and more applicably
(What I wrote might not have done it justice, but the idea remains)

kaikalii: I think it should be put up with the rest of the `fix` explanation then.

amatgil:
The off-github conversation lead to the conclusion that a rewrite is better than a tldr section, so this is superceded by https://github.com/uiua-lang/uiua/pull/689

PR #689, opened 2025-04-08
Rewrite :fix: section of the Advanced section

amatgil:
Continues https://github.com/uiua-lang/uiua/pull/688


kaikalii:
I made some changes.

amatgil:
That seems much better, I do agree; the distinction I had between scalar and number-of-rows wasn't good
Is there anything blocking this, then?

kaikalii:
Nope, seems good. I just wanted your feedback.

PR #693, opened 2025-04-12
Add subscripts for `length` and `shape`

Omnikar:
Subscripted `length` gets the length of the Nth axis.
Subscripted `shape` gets the shape from the first N axes.

Side note, why is the internal name for `⧻` `Len` and not `Length`?


kaikalii:
These would also need to be added to [`Primitive::subscript_sig`](https://github.com/uiua-lang/uiua/blob/main/src/primitive/mod.rs#L515) and the [Subscripts page](https://github.com/uiua-lang/uiua/blob/main/site/src/other.rs#L497).
Also, needs tests.

Omnikar:
Where do I put the tests?

kaikalii:
Probably in `/tests/monadic.ua`

PR #702, opened 2025-05-07
Sattolo -> Fisher-Yates

amatgil:


amatgil:
The first paragraph of [this](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Implementation_errors) mentions the exact implementation bug and how to solve it (which is this PR)

PR #705, opened 2025-05-26
add new optimization for sign len path

regr4:
Hey, the person from discord here. I decided to try implementing an optimization for sign len path myself. Hope it's usable :)


src/algorithm/path.rs
```diff
@@ -408,6 +418,11 @@ fn path_impl(
                 return Err(env.error("No path found"));
             }
         }
+        PathMode::Exists => {
+            env.push(Array::scalar(
+                ends.into_iter().next().into_iter().count() as u8
+            ));
```
kaikalii, line 424: I think this can just be
```rust
env.push(!ends.is_empty())
```

PR #725, opened 2025-06-21
make dims_to_flat and i_dims_to_flat accept an iter

alex-s168:


Omnikar:
You should just be able to do `index: impl Iterator<Item = impl Borrow<usize>>` right?

kaikalii:
Functions should generally take `IntoIterator` rather than `Iterator`. Makes it easier to use.

Omnikar:
Also, maybe we should use `IntoIterator` instead of `Iterator` to make it more dynamic?

alex-s168:
> You should just be able to do `index: impl Iterator<Item = impl Borrow<usize>>` right?

Oh true
done

Omnikar:
Oh lol my end didn't update with Kai's comment before mine was posted.

alex-s168:
> Functions should generally take `IntoIterator` rather than `Iterator`. Makes it easier to use.

But do iterators implement `IntoIterator`?

Omnikar:
Yes they do

kaikalii:
Yes

Omnikar:
I'm pretty sure that if you change it to use `IntoIterator` then this PR doesn't even have to touch any of the call sites.
As in, changing `&x` to `x.iter()` is no longer necessary.

alex-s168:
interesting

done

Omnikar (approved):

ekgame (approved):

PR #731, opened 2025-06-21
fix private macros beeing 'private unused' in the LSP when they are used

alex-s168:


alex-s168:
it doesn't fix the issue in all cases, but idk

Marcos-cat:
What cases are still unhandled?

alex-s168:
> What cases are still unhandled?

image

ekgame (approved):

PR #738, opened 2025-06-22
Add &seek to changelog

alex-s168:


Omnikar (commented):
Maybe it should go above the removal bullets alongside the other "Add..." bullets? Also, I think

parser/src/defs.rs
```diff
@@ -3739,7 +3739,11 @@ sys_op! {
     /// ex: &cl &w "Hello, world!" . &fc "file.txt"
     ///   : &fras "file.txt"
     (2(0), Write, Stream, "&w", "write", Mutating),
-    /// Moves to a absolute position in the file under the stream.
+    /// Moves to an absolute position in the file under the stream.
```
Omnikar, line 3742: This wording is kind of confusing. Does this function only work with file streams, or does it also work with other streams?

alex-s168: well it currently only works in file streams, and I don't think any other streams will support it any time soon

how should I word it?

Omnikar: How's "Move to an absolute position in a file stream"?

Omnikar: That also matches the imperative structure and lack of ending punctuation of other primitives' docs pages.

parser/src/defs.rs
```diff
@@ -3739,7 +3739,11 @@ sys_op! {
     /// ex: &cl &w "Hello, world!" . &fc "file.txt"
     ///   : &fras "file.txt"
     (2(0), Write, Stream, "&w", "write", Mutating),
-    /// Moves to a absolute position in the file under the stream.
+    /// Moves to an absolute position in a file stream
```
Omnikar, line 3742: Sorry to be this nitpicky, but other docs pages are generally imperative, not declarative.
```suggestion
    /// Move to an absolute position in a file stream
```

amatgil (dismissed):

Omnikar:
Wait a sec will the fact that the example doesn't work in the pad cause CI to fail?

Omnikar (changes requested):
`&ffi`'s docs use `ex!` because of the intentional failure on the website.

alex-s168:
Maybe, but then let's wait for #739 to be merged first

then it won't fail

parser/src/defs.rs
```diff
@@ -3739,7 +3739,11 @@ sys_op! {
     /// ex: &cl &w "Hello, world!" . &fc "file.txt"
     ///   : &fras "file.txt"
     (2(0), Write, Stream, "&w", "write", Mutating),
-    /// Moves to a absolute position in the file under the stream.
+    /// Move to an absolute position in a file stream
+    ///
+    /// If the position is negative, it is an offset from the file end.
+    ///
+    /// ex: &rs 5 ⊸&seek 10 &fo "example.txt"
```
Omnikar, line 3746:
```suggestion
    /// ex: &rs 4 ⊸&seek 47 &fo "example.txt"
```

amatgil (approved):
If tests pass

Omnikar (approved):

PR #718, opened 2025-06-19
add file seek (&sb)

alex-s168:
https://discord.com/channels/1156339038748413952/1385007087582445690

PR #739, opened 2025-06-22
support &seek in pad

alex-s168:
Use custom `StreamSeek` enum, because `std::io::SeekFrom` allows for more cases than we are using

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -160,6 +160,21 @@ impl fmt::Display for HandleKind {
     }
 }
 
+#[allow(missing_docs)]
+pub enum StreamSeek {
+    Start(usize),
+    End(usize),
+}
+
+impl Into<std::io::SeekFrom> for StreamSeek {
+    fn into(self) -> std::io::SeekFrom {
+        match self {
+            StreamSeek::Start(off) => std::io::SeekFrom::Start(off as u64),
+            StreamSeek::End(off) => std::io::SeekFrom::End(-(off as i64)),
+        }
+    }
+}
+
 #[cfg(feature = "image")]
 pub(crate) type WebcamImage = image::RgbImage;
 #[cfg(not(feature = "image"))]
@@ -299,7 +314,7 @@ pub trait SysBackend: Any + Send + Sync + 'static {
         Err("Writing to streams is not supported in this environment".into())
     }
     /// Go to an absolute file position
-    fn seek(&self, handle: Handle, offset: std::io::SeekFrom) -> Result<(), String> {
+    fn seek(&self, handle: Handle, offset: StreamSeek) -> Result<(), String> {
         Err("Seeking streams is not supported in this environment".into())
     }
     /// Create a file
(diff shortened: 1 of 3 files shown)


amatgil:
Perchance tests could get added wrt &seek?

alex-s168:
> Perchance tests could get added wrt &seek?

well kai accepted #718 , and it doesn't have tests either.

could add in the future maybe but idk

src/sys/mod.rs
```diff
@@ -160,6 +160,21 @@ impl fmt::Display for HandleKind {
     }
 }
 
+#[allow(missing_docs)]
```
amatgil: This should definitely not be here

amatgil:
> > Perchance tests could get added wrt &seek?
> 
> well kai accepted #718 , and it doesn't have tests either.
> 
> could add in the future maybe but idk

Fair enough

src/sys/mod.rs
```diff
@@ -160,6 +160,21 @@ impl fmt::Display for HandleKind {
...
+pub enum StreamSeek {
+    Start(usize),
+    End(usize),
+}
```
Omnikar, line 167:
```suggestion
/// Reference point and offset to seek a position in a stream with
pub enum StreamSeek {
    /// Seek a position forward from the start of the stream
    Start(usize),
    /// Seek a position backward from the end of the stream
    End(usize),
}
```

src/sys/mod.rs
```diff
@@ -160,6 +160,23 @@ impl fmt::Display for HandleKind {
...
+
+impl Into<std::io::SeekFrom> for StreamSeek {
+    fn into(self) -> std::io::SeekFrom {
+        match self {
+            StreamSeek::Start(off) => std::io::SeekFrom::Start(off as u64),
+            StreamSeek::End(off) => std::io::SeekFrom::End(-(off as i64)),
+        }
+    }
+}
```
Omnikar, line 178:
```suggestion
impl From<StreamSeek> for std::io::SeekFrom {
    fn from(value: StreamSeek) -> Self {
        match value {
            StreamSeek::Start(off) => std::io::SeekFrom::Start(off as u64),
            StreamSeek::End(off) => std::io::SeekFrom::End(-(off as i64)),
        }
    }
}
```
Clippy says to implement `From` instead of `Into` because the former gives you the latter but not the other way around.

src/sys/mod.rs
```diff
@@ -160,6 +160,23 @@ impl fmt::Display for HandleKind {
     }
 }
 
+/// Reference point and offset to seek a position in a stream with
```
Omnikar, line 163:
```suggestion
/// Reference point and offset to seek a position in a stream with
/// This is used instead of `std::io::SeekFrom` because the latter is more general than what Uiua uses, so using a more specific enum reduces checks elsewhere in the code.
```

amatgil (approved):

Omnikar (approved):

amatgil (approved):

PR #741, opened 2025-06-22
Fix FFI tests

magistau:
This effectively makes `&ffi` leak all allocated memory. It should probably be possible to only leak the necessary data, but the existing FFI code is a bit too convoluted to comprehend.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -222,7 +222,7 @@ pub(crate) use enabled::*;
 mod enabled {
     use std::{
         any::{type_name, Any},
-        mem::{forget, take, transmute},
+        mem::{forget, take, transmute, ManuallyDrop},
         ptr, slice,
     };
 
@@ -639,10 +639,10 @@ mod enabled {
 
     type ListStorage<T> = (*mut T, Box<[T]>);
 
-    #[derive(Default)]
+    #[derive(Default, Debug)]
     struct FfiBindings {
-        arg_data: Vec<Box<dyn Any>>,
-        other_data: Vec<Box<dyn Any>>,
+        arg_data: Vec<ManuallyDrop<Box<dyn Any>>>,
+        other_data: Vec<ManuallyDrop<Box<dyn Any>>>,
         args: Vec<Arg>,
     }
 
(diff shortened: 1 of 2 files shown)


magistau:
For completeness, here's a helpful GDB script:
```gdb
break src/ffi.rs:630
  commands
    disable $_hit_bpnum
  end
set $drop_break = $bpnum
disable $drop_break
tcatch load ffi_lib
  commands
    silent
    break dummy_md5 if out != 0
      commands
        set $msg=&(*m@len)
        set $out=&(*out@len)
        printf "Updated $msg and $out\n"
        enable $drop_break
        display
      end
    continue
  end
display $out
```
Run with `rust-gdb -x ffi-brp.gdb --args target/debug/uiua run tests_ffi/test.ua target/debug/libffi_lib.so` after building both `uiua` and `tests_ffi`

src/ffi.rs
```diff
@@ -775,47 +736,45 @@ mod enabled {
...
                 .map(|t| {
                     dbgln!("  exact type");
                     (t, None)
                 })
                 .or_else(|| {
-                    any.downcast_ref::<(*mut T, Box<T>)>().map(|(p, _)| {
+                    self.try_get_as::<(*mut T, Box<T>)>(index).map(|(p, _)| {
                         dbgln!("  ptr type");
-                        (unsafe { &**p }, Some(*p))
+                        // SAFETY: TODO
```
alex-s168, line 747: resolve this todo

alex-s168 (changes requested):

src/ffi.rs
```diff
@@ -652,118 +652,79 @@ mod enabled {
...
+            let ptr = Box::into_raw(arg.into_bytes_with_nul().into_boxed_slice());
+            // SAFETY: the pointer was created using `Box::into_raw`
+            // and `[i8]` is compatible with `[u8]`
+            let arg = unsafe { Box::from_raw(ptr as *mut [c_char]) };
+            self.push_list(arg)
         }
-        fn push_list<T: Any + 'static>(&mut self, mut arg: Box<[T]>) -> *mut T {
+        fn push_list<T: Any>(&mut self, mut arg: Box<[T]>) -> *mut T {
             let ptr = if arg.is_empty() {
+                // TODO: is this special case necessary?
```
alex-s168, line 700: resolve this TODO too

src/ffi.rs
```diff
@@ -639,10 +639,10 @@ mod enabled {
 
     type ListStorage<T> = (*mut T, Box<[T]>);
 
-    #[derive(Default)]
+    #[derive(Default, Debug)]
```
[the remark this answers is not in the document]
magistau: Well, it doesn't hurt to have a derived Debug I think? I mean, I can remove it, but I don't see a problem either way.

src/ffi.rs
magistau: I think this is impossible without input from Kai — I have no idea about why a null pointer is used there, and I'm not sure where the code dependent on this could be.

src/ffi.rs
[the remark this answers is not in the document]
Omnikar: Yeah I don't see a reason not to derive `Debug` if it's possible to.

Omnikar:
I definitely think this PR (and #742) need input from Kai before we can think about merging them. I personally very much don't know enough about what's going on in this PR to be able to comment on it at the moment unfortunately.

kaikalii:
How does this fix tests?

magistau:
1. On macOS, a different file extension is used for dynamic libraries
2. `&ffi` doesn't deallocate data before returning it anymore. Tests were failing because `&ffi` implementation actually contained a bug causing UB — namely, an output buffer would be deallocated before `&ffi` even returns anything.

kaikalii:
This should just be fixed in `&ffi` then. Not by making it always leak.

kaikalii:
What's the test case that was causing issues?

magistau:
> This should just be fixed in `&ffi` then. Not by making it always leak.

Well, the better solution would be leaking just the needed data, but I do not understand Uiua code well enough for that. As noted in the PR description:
> It should probably be possible to only leak the necessary data, but the existing FFI code is a bit too convoluted to comprehend.

To be honest, after spending >24h studying the code, running `gdb` over and over, digging into `libffi` and others, I still have little idea about what exactly is `FfiBindings` for, especially the `other_data` field. I rewrote a part of it to be as clear as possible, and there are two things I fail to understand in that part (see the TODOs). There's also some `value.meta.pointer` stuff and five macro declarations in a single 400SLoC `do_ffi` function, which is too intimidating to understand.

magistau:
> What's the test case that was causing issues?

Seemingly the only line testing output buffers: https://github.com/uiua-lang/uiua/blob/8a9de2dccf4dbb0dc896285c79c7ba6d7e0c770e/tests_ffi/test.ua#L41

kaikalii:
`FfiBindings` stores data that is being passed to foreign functions. libffi requires you to pass the data as pointers, and that data may *be* pointers, so it has to be allocated on the Rust side somewhere. Those allocations are meant to last until the foreign function call is complete.

magistau:
> Those allocations are meant to last until the foreign function call is complete.

Well, this is not the cause for output buffers, since those are supposed to live after `&ffi` is done.

kaikalii:
Would this make you have to use `&free` all the time?

magistau:
Well, it's not like you _have_ to, but yeah, that's what happens.

Marcos-cat:
This is solved by out parameter pointers in #783

PR #742, opened 2025-06-22
ci: test more generously

magistau:


.github/workflows/build_latest.yml
```diff
@@ -26,32 +26,6 @@ concurrency:
...
-      with:
-        path: target
-        key: build-${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
-        restore-keys: |
-          build-${{ runner.os }}-
-    - name: Run site tests
-      run: cargo test -p site
   # Build the site
   build_site:
     needs: test_site
```
alex-s168, line 31: how does that work?
you removed `test_site`

.github/workflows/validate.yml
```diff
@@ -1,33 +1,8 @@
 name: Validate
 
 on:
-  push:
-    branches: ["main"]
-    paths:
```
alex-s168, line 6: was there maybe a reason why this was restricted to paths?
I don't know

.github/workflows/validate.yml
```diff
@@ -1,33 +1,8 @@
...
-      - "src/**"
-      - "parser/**"
-      - "site/**"
-      - "pad/**"
-      - "tests*/**"
-      - "changelog.md"
-      - "readme.md"
-      - ".github/workflows/validate.yml"
-
+  - push
```
alex-s168, line 4: wouldn't this then also run on the `site` branch?

alex-s168 (changes requested):

.github/workflows/build_latest.yml
magistau: `--workspace` is used instead

alex-s168: but the line says `needs: test_site`

I don't know much about github actions, but it sounds a lot like it needs `test_site`

.github/workflows/validate.yml
magistau: Probably in order to not run tests when they are not necessary? Not sure.

.github/workflows/validate.yml
magistau: It could probably be excluded then?

.github/workflows/validate.yml
alex-s168: meh it's probably fine?

.github/workflows/validate.yml
alex-s168: idk if that's possible in GH actions, but yes it should be

.github/workflows/build_latest.yml
magistau: Right, didn't notice this.

magistau:
CI failing is due to FFI tests currently being broken (see #741).

alex-s168 (approved):
seems fine

we'll see if it works after merging

kaikalii:
CI passes for `main`. Why not this?

magistau:
> CI passes for `main`. Why not this?

CI on `main` doesn't run tests in `tests_ffi`, CI in this PR does, as I [mentioned in this thread](https://github.com/uiua-lang/uiua/pull/742#issuecomment-2994572196)

PR #743, opened 2025-06-23
Dependency cleanup

magistau:


amatgil:
I'm surprised the site compiles without `getrandom` as a dependency (for the rand crate)

PR #744, opened 2025-06-23
make removing # Experimental! easier

lemonlime4:
- Add the shortcut ctrl+shift+E
- Make the 🧪 button remove experimental comment if it's there
- Make ctrl+E (and 🧪) more consistent

### DIFF
--- a/pad/editor/src/lib.rs
+++ b/pad/editor/src/lib.rs
@@ -484,8 +484,7 @@ pub fn Editor<'a>(
                     Cursor::Ignore
                 };
                 state.set_code(new_code, cursor);
-            }
-            else if code == "# Experimental!" {
+            } else if code == "# Experimental!" {
                 state.set_code("", Cursor::Set(0, 0));
             }
         });
@@ -1021,8 +1020,7 @@ pub fn Editor<'a>(
         let code = get_code();
         if code.starts_with("# Experimental!\n") || code == "# Experimental!" {
             remove_experimental();
-        }
-        else {
+        } else {
             insert_experimental();
         }
     };
--- a/pad/editor/src/lib.rs
+++ b/pad/editor/src/lib.rs
@@ -1017,12 +1017,20 @@ pub fn Editor<'a>(
         }
     };
 
-    let on_insert_experimental = move |_| insert_experimental();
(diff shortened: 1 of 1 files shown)


pad/editor/src/lib.rs
```diff
@@ -460,23 +460,36 @@ pub fn Editor<'a>(
     let insert_experimental = move || {
         state.update(|state| {
             let code = get_code();
-            if code.starts_with("# Experimental!") {
+            if code.starts_with("# Experimental!\n") || code == "# Experimental!" {
```
Omnikar, line 463: Why not just check if `code.starts_with("# Experimental!")`?

pad/editor/src/lib.rs
```diff
@@ -744,6 +757,8 @@ pub fn Editor<'a>(
             "z" if os_ctrl(event) => state.update(|state| state.undo()),
             // Insert # Experimental! comment
             "e" if os_ctrl(event) => insert_experimental(),
+            // Remove # Experimental! comment
+            "E" if os_ctrl(event) => remove_experimental(),
```
Omnikar, line 761:
```suggestion
            // Remove # Experimental! comment
            "e" | "E" if os_ctrl(event) && event.shift_key() => remove_experimental(),
            // Insert # Experimental! comment
            "e" if os_ctrl(event) => insert_experimental(),
```
At least on my machine, the ⇧⌘E input was not triggering the functionality without this change. The new match branch must be above the old one, since the latter catches more cases. See the undo/redo bindings above.

pad/editor/src/lib.rs
```diff
@@ -1001,12 +1016,19 @@ pub fn Editor<'a>(
...
+        } else {
+            insert_experimental();
+        }
+    };
+    let toggle_experimental_button = view! {
         <button
             class="info-button"
-            data-title="Add # Experimental"
-            on:click=on_insert_experimental
+            data-title="Toggle # Experimental"
```
Omnikar, line 1030:
```suggestion
            data-title="Toggle # Experimental!"
```

Omnikar (approved):
I guess technically this isn't a bugfix, but I don't see any reason to be against this change. Code looks good to me.

PR #745, opened 2025-06-26
Rational number implementation

amatgil:
Big rationals implementation (very wip as of now)


amatgil:
I wonder if `RealArrayValue: Copy` is necessary; a rational should probably implement it but in this impl they're on heap

PR #750, opened 2025-06-30
thats just a bug waiting to happen

Tux9001:


Marcos-cat:
Nuh uh, its just a funny

kaikalii:
Ya I'm leaning toward keeping this. Uiua is supposed to be a fun language that makes you smile, and little things like this do that for me.

That said, the place where this is more likely to cause issues is in some macro, as it requires you to literally write "22/7" in your code.

Omnikar (approved):
Humorous sure, but a joke shouldn't make the language intentionally bugged. I agree with removing this.

amatgil (changes requested):
I think it should stay, it's very haha-funny and incredibly specific

Perchance it could be changed to 355/113, but I'd vote for leaving it as is

Omnikar:
It still makes the language unsound. And think of it from the perspective of someone new to the language. If this aspect of math is intentionally wrong, what else is? It makes it harder to take the language seriously as a programming language. People already often see Uiua as an esolang when they first find out about it. This doesn't help.

Omnikar:
Easter eggs are great when they don't affect the functionality of the language, like the `# Boo!` semantic comment or the "Why would you need that many dimensions?" error message with `geo`.

kaikalii:
I actually agree now.

PR #767, opened 2025-07-08
Extending the available sizes for ffi structs

amatgil:


amatgil:
The sizes are written out manually, for now
I think it should be possible to generate the sizes from a range, perhaps using quote?

kaikalii:
I worry that adding this many inflates the binary size.

amatgil:
> I worry that adding this many inflates the binary size.

The difference is 0.4MB, it seems
(I compiled it both ways and compared sizes)

kaikalii:
With release optimizations? That's kind of a lot.

amatgil:
It's 1.1% of the binary size, is that a lot? /genq

Marcos-cat:
This is solved by #783

PR #776, opened 2025-07-15
fix: fork and fix in planet notation shorthand

magistau:
For some reason, planet notation only worked with strings of following forms:
- `fp`
- `f[gd]+[ip]`
- `[gd]{2,}`
- `d+f`

While the following did not work:
- `fi`
- `f[gd]+`
- `f[gd]*f`
- `[gd]*g[gd]*f`

Although it's a bit harder to describe using just regex — for instance, `ggdf` would parse, but as two juxtaposed shorthands, like `gg df`.

With this change, planet notation is treated as described by [Planet aliases](https://www.uiua.org/docs/idioms#planet-aliases): that is, `f?[gd]*[fip]?` containing at least two characters.

### DIFF
--- a/parser/src/split.rs
+++ b/parser/src/split.rs
@@ -263,40 +263,21 @@ pub fn split_name(name: &str) -> Option<Vec<(PrimComponent, &str)>> {
                 continue;
             }
             // 1-letter planet notation
-            if sub_name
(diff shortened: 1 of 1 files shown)


kaikalii:
Thank you! I should make some tests for this.

PR #777, opened 2025-07-15
feat: glyph and subscripts for `dump`

magistau:
Uses `‽`, can be formatted from `?!`. Just like subscripted `stack` can be formatted from a sequence of `?`s, subscripted `dump` can be formatted from `?!` followed by a sequence of `?`s.


kaikalii:
I don't want `dump` to have a glyph.

magistau:
> I don't want `dump` to have a glyph.

Why not?
Also, what about subscripts?

kaikalii:
It's simply not used enough. I think glyphs should only be for stuff that's kind of fundamental.

magistau:
> It's simply not used enough. I think glyphs should only be for stuff that's kind of fundamental.

Well, I can say for sure that I used `dump` way more than `base`, `fall`, `scan`, `find` and maybe others. Hard to compute exact statistics since `stack` and `dump` usually only exist while iterating on the code and don't get into the finished program. And I faced the lack of `dump` subscripts several times.

PR #782, opened 2025-07-18
fix: use stderr for diagnostic data

magistau:


src/sys/native.rs
```diff
@@ -1257,7 +1255,7 @@ pub fn print_stack(stack: &[Value], color: bool) {
     }
     if stack.len() == 1 || !color {
         for value in stack {
-            println!("{}", value.show()); // Allow println
+            eprintln!("{}", value.show());
```
kaikalii, line 1258: These are the output of the program, so maybe these should still be stdout?
I guess "output" is relative.

kaikalii:
I'm happy to merge once conflicts are resolved.

magistau:
Resolved by dropping changes in `src/ffi.rs`, since `dbgln!` doesn't exist there anymore.

PR #785, opened 2025-07-22
Add `parse` subscripts

Marcos-cat:


kaikalii:
Will you please add some tests for this?

Marcos-cat:
Definitely

PR #788, opened 2025-07-23
feat: add support for UDP sockets

donstenzel:
I wanted to work with UDP sockets and looked into the implementation in Rust's standard library. The API provided is different to TCP streams / listeners, in that there is no persistent connection when using UDP sockets. As such, special functions are required to work with them.

I have added four functions:
- `&udpb` binds a socket to an address
- `&udpr` receives a datagram in bytes on a socket
- `&udps` sends a datagram to an address on a socket
- `&udpsml` changes the maximum message length of a socket

The message length is required for receiving messages. It is used to initialize the buffer to which the datagram is written.
I decided to keep this state internally, as the value is unlikely to be changed often. This way, users don't have to provide the length every time a message is received.

I have tested the functions by sending datagrams to myself using multiple terminal sessions. I have also tested a simple pooled ping pong server, which worked.


kaikalii:
This is awesome, thank you!

donstenzel:
Happy to help out wherever I can :)

kaikalii:
Wait, what is the purpose of the maximum message length?

donstenzel:
When receiving a message, you have to initialize a buffer to some size, if messages are longer they will be chopped off. The theoretical maximum size for a single datagram is about 65kB, which would be unnecessarily large for many applications.

PR #794, opened 2025-07-26
Add `each` alias for `rows,0`

Omnikar:
I'm not familiar with the `Sequence` trait but it looks like it supports `i16` but not `i32`.

### DIFF
--- a/parser/src/split.rs
+++ b/parser/src/split.rs
@@ -56,6 +56,8 @@ pub enum PrimComponent {
     Prim(Primitive),
     /// A numeric component
     Num(NumComponent),
+    /// Subscript 0
+    Sub0,
     /// Subscript 2
     Sub2,
 }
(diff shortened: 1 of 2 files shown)


parser/src/split.rs
```diff
@@ -72,13 +73,15 @@ impl From<NumComponent> for PrimComponent {
...
+const SUBSCRIPT_DIGIT_STRS: [&str; 10] = ["₀", "₁", "₂", "₃", "₄", "₅", "₆", "₇", "₈", "₉"];
+
 impl PrimComponent {
     /// Get the name of this component
     pub fn name(&self) -> &'static str {
         match self {
             PrimComponent::Prim(prim) => prim.name(),
             PrimComponent::Num(num) => num.name(),
-            PrimComponent::Sub2 => "₂",
+            PrimComponent::SubN(n) => SUBSCRIPT_DIGIT_STRS[*n as usize],
```
kaikalii, line 84: I think I'd prefer the hardcoded approach. The way you have here adds unused data in the constant.

PR #809, opened 2025-08-04
feat: under support for regex

magistau:


kaikalii:
Can you add some tests for this? Probably in `tests/under.ua`. Probably a good number of them, since it's a complicated feature.

magistau:
> Can you add some tests for this? Probably in `tests/under.ua`. Probably a good number of them, since it's a complicated feature.

Done

kaikalii:
Awesome, thank you!

PR #813, opened 2025-08-07
Allow prime in binding names

Omnikar:
Currently, something like `F′₁` works, but `F₁′` does not.


kaikalii:
Can you add a note about this to the bindings tutorial?

Marcos-cat:
I think the subscript should stay after the prime, to cause less problems in a future with custom subscripts

Omnikar:
> Can you add a note about this to the bindings tutorial?

Done

PR #823, opened 2025-08-13
Change `⊢⇌` to `⊣` in optimizations tests

Jacob-Lockwood:
The documentation aspect of this was addressed in [commit 5dba9c9](https://github.com/uiua-lang/uiua/commit/5dba9c91a03faa1479ba65d5e7c81f31ff5f3770), but the optimizations tests were never changed to reflect it.


kaikalii:
It's like that on purpose. The first rev to last optimization is also being tested.

Jacob-Lockwood:
Maybe it would be better to add a separate test for that, then.

Omnikar:
It's not a separate test. `first reverse rise` and friends are their own optimizations.

Jacob-Lockwood:
Interesting. I wonder if that's a good thing? Still, I'll close the PR.

PR #839, opened 2025-09-14
"Fix" bits,n when n>127

amatgil:
Right now, when too many bits are requested, the result is incorrect. Since the function already errors when values are greater than u128::MAX, i made the function return a uiua error when the number of bits is itself also greater than 127.

(I've also added related tests, but they might not be sufficient)


amatgil:
Another idea would be to pad with the corresponding number of zeros to the right
I can change this PR to do that, it'll likely be more useful

amatgil:
I have changed it to be padding instead of erroring

Issue #849, opened 2025-10-01
`⬚ _ \` (fill scan) is inconsistent on empty arrays


PistachioCake:
[Pad with context](https://www.uiua.org/pad?src=0_18_0-dev_1__eJyL0VaIjuV6tGaWYQyYxRWjrWAYbxRvDBODcADxmguZ).

Note that only the second line outputs a scalar instead of an array, and all the others output an array of the same length as the input. This seems to me like a bug.

I would expect `⬚1\+ []` to evaluate to `[]` instead of `1`, which would align with the other behavior of both `scan` and `fill _ scan`.

PR #850, opened 2025-10-01
fix inconsistency between scan and fill scan

PistachioCake:
Fixes https://github.com/uiua-lang/uiua/issues/849.


kaikalii:
Awesome, thank you!

PR #852, opened 2025-10-02
Adding option to provide folder as test cases destination

tsych0:
Now supporting

`sh
uiua test tests/
`


src/main.rs
```diff
@@ -292,35 +292,51 @@ fn main() {
...
+                let config = FormatConfig::from_source(
+                    formatter_options.format_config_source.clone(),
+                    Some(&path),
+                )
+                .unwrap_or_else(fail);
+                format_file(&path, &config).unwrap_or_else(fail);
+                let mut rt = Uiua::with_native_sys()
+                    .with_file_path(&path)
+                    .with_args(args.clone());
+                println!("Testing {}:", path.display());
```
kaikalii, line 327: This should be `eprintln!`.

kaikalii:
This is great, thanks! Can you please also add an entry to the changelog under `# 0.18.0` `## Interpreter` section?

PR #855, opened 2025-10-15
rescale domain coloring so that magnitude 1 is perfect hue

ElNico56:


kaikalii:
Will you fix the clippy lint?

ElNico56:
ye

PR #865, opened 2025-11-03
Remove with and off promotion and allow on monadic functions

Omnikar:


kaikalii:
Can you add a style diagnostic when used on monadic functions?

`self.emit_diagnostic(...)`

PR #866, opened 2025-11-03
Iterative optimization pass

ndren:
Uiua programs with many nested AST nodes can slow down the recursive optimization pass, for example:
```
# Experimental!
quote(
  "∘"
  ↯ 15
  /$"_|_"
  $"fork[_]"
)1
```
This code changes the optimization pass to iteratively optimize the nodes by explicitly keeping a stack rather than recursing, allowing the code above to run until the compiler itself rejects it rather than the optimizer (around N=120). 

Unfortunately this code uses unsafe since the borrow checker can't recognise the data is changed in-place, is this something that can be avoided? Also, is there neater tree traversal pattern that would be lead to less complex code?

### DIFF
--- a/src/compile/optimize.rs
+++ b/src/compile/optimize.rs
@@ -27,76 +27,167 @@ impl Node {
         self.optimize_impl(OptLevel::Full, true)
     }
     fn optimize_impl(&mut self, level: OptLevel, opt_single: bool) -> bool {
-        let mut optimized = false;
-        fn optimize_run(nodes: &mut EcoVec<Node>, level: OptLevel, opt_single: bool) -> bool {
-            let mut optimized = false;
-            for node in nodes.make_mut() {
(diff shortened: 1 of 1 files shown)


kaikalii:
This is just unrolling the recursion right? I'm not convinced this requires `unsafe`.

ndren:
> This is just unrolling the recursion right?

Yeah that's all this does
> I'm not convinced this requires unsafe.

I don't think it requires unsafe, but anything I've tried would complain about mutable reborrowing in different arms of the node optimization, even though they are unrelated to each other. I was wondering if there's a neater way without cloning the entire thing?

kaikalii:
00865231939b3141677368c allows `fork` to take an arbitrary number of non-nested functions. This solves this particular instance of this issue.

ndren:
Awesome, thanks! The other cases are probably fine to handle with sensible recursion limits.

PR #868, opened 2025-11-07
feat: Add some additional shadowable constants

Bazinga9000:
This PR adds a few shadowable constants, perhaps of some use:

**Math:**
- `φ`, the golden ratio.
- `γ`, the Euler-Mascheroni constant.
(Both of these constants are in the process of being added as rust constants `std::f64::consts::PHI` and `std::f64::consts::EGAMMA`, but that's not stabilized yet (relevant feature flag is `more_float_constants`) so this PR just uses the raw definitions proposed there.)

**Fun:**
- `PlanetSymbols` and `ZodiacSymbols`, the astronomical symbols corresponding to existing constants `Planets` and `Zodiac`.
- `Elements` and `ElementSymbols`, the names and chemical symbols of the 118 chemical elements.


kaikalii:
Can you add to the changelog?

Bazinga9000:
Yes, done!

Bazinga9000:
Clippy failed with excessive precision on the math constants, so I truncated them as suggested.

Marcos-cat:
In the changelog the constant name is referred to as `Element` but in the code it is `Elements` (I think the plural is more appropriate)

PR #987, opened 2026-04-11
Use exp() for exponential

ndren:
This PR replaces the uses of generic pow for a more accurate and faster usage of exp, while preserving equality between complex and real inputs. That is, this will still return true:
```
/×=∩ₑ⊸(ℂ0)⇡ 1e6
```

Real performance:
```
⊙◌⍜now ₑ ⇡ 1e6
```
0.02626204490661621 -> 0.013309240341186523

Complex performance:
```
⊙◌⍜nowₑ˙⊞ℂ ⇡ 1e3
```
0.13152217864990234 -> 0.06182527542114258

I will note some inputs now return different values, but this is either because it's more accurate (`ₑ3` 20.085536923187664 -> 20.085536923187668) or gets rid of an incorrect NaN result (`ₑ∞r0i` ¯NaN+NaNi -> ∞ℂ)

### DIFF
--- a/parser/src/complex.rs
+++ b/parser/src/complex.rs
@@ -1,6 +1,6 @@
 //! The [`Complex`] type
 
-use std::{f64::consts::E, fmt, ops::*};
+use std::{fmt, ops::*};
 
 use bytemuck::{Pod, Zeroable};
 use serde::*;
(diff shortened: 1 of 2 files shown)


kaikalii:
Ooh, very good, thanks!

PR #989, opened 2026-04-25
[NOSQUASH] Optimize assets

ndren:
This PR optimizes assets with different lossless compression tools:
```
oxipng -o6 *.png
jpegtran -optimize lena.jpg lena-new.jpg
gifsicle --optimize=3 bad-apple.gif -o bad-apple-new.gif

```
~cwebp -exact elevation.webp -o elevation2.webp~ Not useful, see review below.
Overall diff:
```
  23 binary files changed 11367613 b (10 Mb) -> 6969544 b (6 Mb) => [-4398069 b (-4 Mb)]
```
This should help the site/assets load more quickly over a slow connection (including the include_bytes! used in the wasm blob)

If this PR is up to scratch (I have gone through and checked all the assets pass the eye test) I would recommend not squashing this PR so that if there are any issues it's easy to revert the individual commit.


src/assets/cats.webp
kaikalii, line 1: I definitely seems like fidelity is lost in this one.

ndren: Oops, thank you for mentioning, it turns out
`cwebp -exact` does not do as intended and the option is ignored, it requires `cwebp -lossless -exact ` which makes the file bigger.
I've pushed back previous commit so that the repo won't keep both copies.

kaikalii:
Awesome!

PR #993, opened 2026-08-04
Fix multivector compile errors w/ --no-default-features

ishaan-sen:
I get a few errors pertaining to multivectors when compiling w/o default features. It seems that they're all resolvable by gating things behind  `#[cfg(feature = "ga")]`; I've done this and it seems to work and pass all the tests.


kaikalii:
Uiua can run on embedded targets?

ishaan-sen:
it turns out that it can

ishaan-sen:
(force push was to open a new pr, #994, for the separate portable-atomic issue)

Issue #1002, opened 2026-08-25
Suggestion: nested unders


UnrelatedString:
I was just writing some string manipulation code for funsies, and to my dismay discovered that I could not make a regex substitution preserve the case of only the first character with `⍜⍜⊢⌵`. Although I imagine the implementation may be non-trivial, it seems sensible to me that `⍜⊢⌵` should have an under inverse in precisely the same way as `⌵` itself does, saving the same context to be re-applied within the inner under.

...actually started trying to work on an implementation, and this might really not generalize well given the under inverse being applied after the first half of the under would need to consume its context from under the inner function's, or else use a stack of separate under stacks... an expandable tape of under stacks? My brain hurts trying to think about how more than one layer of this would want to work. 

So it's a biiiiit beyond my ability to hack a PR together for, and would require one or several major internal changes, but the idea's still out there, I guess. Maybe certain simple structural unders could even be special-cased into implmods that don't actually use the under stack or something? But that seems like an iffy compromise. I do notice that `Signature` has `under_args` and `under_outputs` which might be usable for effectively dipping each half of the inner function past the inner under, but at this point `SigNode` frankly scares me.

kaikalii:
Can I see the broader code example so I understand how you are trying to use this?

UnrelatedString:
(Not the actual children's language game implemented, nor the exact same code structure, for my own anonymity and not overcomplicating the example.) Wildly impractical, of course, but:

```uiua
$ Quite a Title-Cased Input

⍜regex≡⍚(⍜⊢×⊂"Ub"¯⊃⌵±)"(?i:[aeiou])"
```

where I would have meant to write `(⍜⊢×⊂"Ub"¯⊃⌵±)` as `⍜⍜⊢(¯⌵)(⊂"ub")`. And, likewise, its inverse,

```uiua
$ Qubuubitube uba Tubitlube-Cubasubed Ubinpubut

⍜regex≡⍚(×↘2⊃⌵(±⊢))"(?i:ub[aeiou])"
```

with `⍜⍜⊢(¯⌵)(↘2)`.

UnrelatedString:
...although this does a bad job of motivating it, when you can also just do the whole thing `⍜(¯⌵)` then fold or delete the characters afterwards...

kaikalii:
`⍜⍜` *does* already work but only when the inner `⍜` doesn't require any saved context values.
[...]

PR #1003, opened 2026-08-26
Under for non-from-un unders

UnrelatedString:
See #1002. Completely untested proof of concept dumpster fire with multiple TODOs; ran in several circles before realizing it would not in fact need deep changes to internals if I simply abuse *both* stacks instead.


UnrelatedString:
...or rather, I realized it wouldn't need deep changes to internals if I added a new *modifier* `ImplPrim` for it, but then I realized I don't even need that. After getting 90% of the way to implementing that. Because maybe copy pasting safeguards around the existing only interface to the under stack is not the best idea

UnrelatedString:
The TODO about phoning it in around signatures seems worth addressing, if it's even possible... the motivation being specifically with structural unders, it would feel sad for `⍜⍜⊏` to be left out... like, this is better than nothing, at least, but, huh. (I honestly can't even pretend to guess at understanding how leading values with `Stash` and `MaybeVal` etc. work as is.)

kaikalii:
How does the transformation work, conceptually?

UnrelatedString:
In `⍜⍜f g h`, each half of `⍜g` operates on the under stack as if it were not being done `⍜f`, and the "after" half of `⍜f` then consumes the values supplied by its "before" half from on top of the under stack effects of each individual half of `⍜g`. `⍜f` can be conceptualized as operating on a second under stack, separate from that of `⍜⍜f g` on which each half of `⍜g` operates -- or I guess maybe that's clearer in an example like `⍜(a ⍜z b c)`, where `⍜z b` is simply the (pair of) operation(s) `b`, modified, but operating on the same normal stack and under stack as `a` and `c` in the same sequence of operations, and each consumes precisely its own data from the under stack while being undone just by being done in order; although `z` shares these as well, its effects on the under stack are isolated from `b`, which sees only its effects on its actual arguments on the normal stack, and which does not operate on any of the under stack that pertains to `z`.

kaikalii:
So does it move `⍜g`'s context values to the main stack while `⍜f` gets undone?

UnrelatedString:
The opposite -- `⍜f`'s context values are moved through the main stack before `⍜g`, then retrieved via the main stack afterwards, in order to pass them through `DipN` rather than implement dedicated functionality for dipping in the under stack.

UnrelatedString:
...which is to say, notionally `⍜f`'s context values hang out in the main stack above `⍜g`'s arguments and outputs, but they underlyingly live safely inside a call to `ImplPrimitive::run_mod`.

kaikalii:
Ah okay cool makes sense.

kaikalii:
Would you mind commenting the steps in the code to make it clearer?

UnrelatedString:
Gladly!

kaikalii:
Would you mind adding some tests for this to `tests_special/error.ua` to make sure invalid uses fail properly and don't crash?

kaikalii:
@UnrelatedString

PR #1020, opened 2026-09-07
Swap type-only validate behavior for scalar and box input

Dguto9:
I find the current behavior, where `⊨ℝ` validates both shape and type, unintuitive. A shape provided outside of a box validates only shape, not type; and the system uses box arrays as the method for provision of both type and shape information. 

This change makes non-boxed type information validate only type, and moves the `[0]`-shape validation to `⊨{ℝ}`. Intuitively, this is providing type and shape info, with the shape info empty.


Dguto9:
Still need to fix things in the type system blog post

PR #1028, opened 2026-09-15
Add `unvalidate`

### LINKED ISSUE #1016: Implement `°⊨`(`unvalidate`)
`°⊨`(`unvalidate`) should take an array and return it's type spec and also they array unchanged.

Will require implementing a function that converts a `Type` into a `Value`. The reverse is currently implemented in `Type::from_spec`.

The type spec should always be a box array.

Examples:
```uiua
°⊨ "abc✨"   # [@𝕌│∙3]
°⊨ [1_2 3_4] # [@ℕ│∙2│∙2]
```

JoaoFelipe3:
Should close #1016.


src/compile/invert/un.rs
```diff
@@ -1496,6 +1496,7 @@ inverse!(ImplPrimPat, input, _, ImplPrim(prim, span), {
         UnClip => Prim(Sys(SysOp::Clip), span),
         Retropose => ImplPrim(Retropose, span),
         MvImpl(mode) => ImplPrim(UnMv(mode), span),
+        ValidateImpl(_) => ImplPrim(UnValidate, span),
```
kaikalii, line 1499: You need to add the opposite as well.

JoaoFelipe3: Done

